PR #14, opened 2020-09-02
Translate firmware readme to Chinese.

leegang:
- Translate firmware readme  file to Chinese .
- Typo fix.


Pascal66:
I dont think that putting a translation for each language in the 'root' folder is a good way to do.
What if some need spanish deutch italian or everything else ? An unreadable folder with a tons of translations.

thias15:
Yes, eventually we will just have a basic README in the root folders. All documentation will move to the docs folder and will be served through our website.

thias15:
Thanks @leegang. Could you label PRs in the future, for example with the documentation tag. Also would be nice if you can rebase your fork and try to minimize the number of commits in a PR so it's easier to review.

PR #63, opened 2020-09-28
Remove chmod +x build step

francisduvivier:
The file is already checked into git as executable so this step is not necessary

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -20,7 +20,5 @@ jobs:
       uses: actions/setup-java@v1
       with:
         java-version: 1.8
-    - name: Grant execute permission for gradlew
-      run: chmod +x android/gradlew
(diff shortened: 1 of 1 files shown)


thias15:
As discussed in issue [#65](https://github.com/intel-isl/OpenBot/issues/65), it seems there is more work to do to this build pipeline properly. Do you just want to use this PR to work on these changes?

francisduvivier:
> As discussed in issue [#65](https://github.com/intel-isl/OpenBot/issues/65), it seems there is more work to do to this build pipeline properly. Do you just want to use this PR to work on these changes?

Is good for me, but then there would be an open PR that is not mergeable, is there a standard way to indicate that the PR is (not) ready?

thias15:
Ok, let me just merge it and you can open a new PR. Thanks.

Issue #65, opened 2020-09-29
[Feature] Build and publish Android Debug apk from GitHub Actions 


francisduvivier:
Hi, just first want to say thanks a lot for this project, it really helps me for getting into hobby robotics, AI,  3D printing, electronics and Arduino.
I was looking at the build pipelines for Android and I see that currently, the android app is not being built(since `android/gradlew` is executed from the root of the repo, which has no gradle project) or uploaded. 

I've worked with Bitbucket pipelines and I would love to contribute and get some experience with GitHub's build pipelines, so I would like to improve this in 2 PR's:
1. Add a build step to build the app module, I'm thinking of using https://github.com/marketplace/actions/android-build for that, but I still have to experiment with that a little.  With this, we can already ensure that on every PR, the app build is still working.

2. Upload the debug apk to the build artifacts. I was thinking to follow [this guide](https://github.com/marketplace/actions/upload-files-to-a-github-release) for that, but here also I still have to experiment a little first. With this, people could download this apk to their device so that they would not necessarily need to install android studio. They would still have to change settings on their phone to allow the installation though so this PR should also include some documentation with a (or link to) explanation about installing a debug apk on  your android phone.

A nice third step would be to instead upload a signed apk, but this requires to store the signing keys somewhere safe, so I won't be able to set that up as an external contributor.
And from there, you could also include an automatic deployment to the F-Droid repo or google play store. But also, since it requires signing keys and authorization, it's out of scope for an external contributor I would think.

Let me know if you guys think the above 2 PR's would be useful. The downside will be that the build will take a lot longer, I think it can certainly add multiple minutes to the build.

thias15:
Hi @francisduvivier. 

This is great! 

(1) I never actually checked the workflow, just assumed it worked. So this would be a great contribution! Ideally, it would only run when files in the android folder are changed and not for any push to the repo. 

(2) This is also a good idea. I'm actually just about the tag the first release, now that the project seems to be semi-stable and some people were able to go through the complete process. Just to make sure, this would only run when a new release is tagged correct? There should not be a new release for every push to the repo. 

(3) I also agree on this and we are working on it. However, it requires some internal approvals to host the app in the Playstore.

francisduvivier:
(1) Good point regarding changed files, we can arrange that with https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths, 

[...]

PR #74, opened 2020-10-01
Upload android debug apk on Release

francisduvivier:
The last PR for #65 .
This PR's hooks in to release publish events to automatically upload the android apk for this release. This PR affects the way of working for release, so I understand if you do not want to accept it. But I just wanted to finish my proposed tasks for #65 .

Here is an [example release](https://github.com/francisduvivier/OpenBot/releases/tag/v0.1.1-test) with an automatically uploaded apk, and a the [GitHub Action run](https://github.com/francisduvivier/OpenBot/runs/1194010562?check_suite_focus=true) for that.

Important here is that this will not trigger when you just push a tag, it will only trigger when a release is made in GitHub.
If we would also want it for tag pushes, that is possible, but then it would need to first create the release as well and it would need conditional logic for the case of a tag with no release yet and a manual release in GitHub. Or 1 way of working would need to be chosen.

Since last time a release with notes was created and for keeping it as simple as possible I used the release trigger.

This will trigger on first publication, so if you first pre-release and then release, it would only trigger on the pre-release.

Lastly, the PR also has a small refactor for the gradle.yml for less duplication.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -45,3 +45,4 @@ jobs:
         with:
           name: app-debug.apk
           path: android/app/build/outputs/apk/debug/app-debug.apk
+
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -19,7 +19,9 @@ jobs:
   build:
     name: Gradle Checks
     runs-on: ubuntu-latest
-
+    defaults:
+          run:
+            working-directory: android
     steps:
       - uses: actions/checkout@v2
       - name: Set up JDK 1.8
@@ -33,17 +35,11 @@ jobs:
           key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
           restore-keys: ${{ runner.os }}-gradle
       - name: Style
-        run: |
-          cd android
-          ./gradlew checkStyle
+        run: ./gradlew checkStyle
       - name: Lint
-        run: |
-          cd android
-          ./gradlew lint
+        run: ./gradlew lint
       - name: Build
-        run: |
-          cd android
-          ./gradlew build
+        run: ./gradlew build
       - name: Upload
         uses: actions/upload-artifact@v2
         with:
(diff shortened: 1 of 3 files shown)


thias15:
> The build of the android apk in this PR is because I included '.github/workflows/gradle.yml' in the paths, I figured that if the job definition is changed, it should run also, just to show that it is still working, but I could remove that.

Ok, that makes sense. I did not realize. You are right, changes to the gradle.yml should also trigger a rebuild. 

> Adding the release version and renaming is no problem, I'll add that.

As for the naming, for the source code, Github chooses the following naming convention: 
<repo-name>-<version_number>.<extension>
So let's follow the same naming convention (e.g. OpenBot-0.1.0.apk). I just updated the previous release accordingly.

francisduvivier:
> One more thing. Since this is not a personal repo, but an org repo, the standard `secrets.GITHUB_TOKEN` may not work. I faced this issue with some other actions that sort the incoming issues. You can use `secrets.GH_TOKEN` instead.
> 
> I also learned that some actions (e.g. when people push from forks) do not work at all. See [this commit](https://github.com/intel-isl/OpenBot/commit/c1496c33555d5ecb1d91a16aa928b0a2336a7b99).
> 
> Maybe you have some ideas, since you are much more experienced with actions.

Interesting about the secrets with  org repos, I'll change that to secrets.GH_TOKEN or make it conditional so it still works in forks.

And actually I don't have that much experience with Actions, I have experience with bitbucket pipelines, which is kind of similar. 
But working on these PR's helped a lot for sure! Also not that much experience with pipelines with PR's that come from forks since I've been working mostly on private repo's. But I might have some idea's, I'll check it([this commit](https://github.com/intel-isl/OpenBot/commit/c1496c33555d5ecb1d91a16aa928b0a2336a7b99)) out.

francisduvivier:
Oh please note that I also updated the android README after creating the PR. So check if you're ok with that.

francisduvivier:
Ok, I think I'm done, last sample release here: 
https://github.com/francisduvivier/OpenBot/releases/tag/v0.1.12 (different branch name but same commit)

francisduvivier:
Just to make sure, I also created a fork in an organisation account, merged the PR there, and created a [tag](https://github.com/DuviCom/OpenBot/releases/tag/v0.1.1-test) to test running the [action](https://github.com/DuviCom/OpenBot/runs/1198042386?check_suite_focus=true).

Looks ok there.

Issue #82, opened 2020-10-13
Update Chinese README files.


thias15:
There have been quite a few changed to the README files. @leegang @zhaoyshine if any of you would like to update them, it would be very much appreciated.

leegang:
WIP.

leegang:
#84

PR #84, opened 2020-10-15
Update Chinese README files.

leegang:
Update Chinese README files.


leegang:
#82

thias15:
This is great! Just two minor comments. I noticed for example in the cad folder die the Chinese README is called README copy.md so the linking does not work. Also the links inside the Chinese doc should link directly to the Chinese version README files. Could you fix this please. If you just push to your branch, the PR is updated automatically. Thanks for your contribution!

leegang:
> This is great! Just two minor comments. I noticed for example in the cad folders the Chinese README is called README copy.md so the linking does not work. Also the links inside the Chinese doc should link directly to the Chinese version README files. Could you fix this please. If you just push to your branch, the PR is updated automatically. Thanks for your contribution!

OK, will fix them.

thias15:
Thanks a lot.

PR #101, opened 2020-11-10
1st test adding a battery status field -graphical only- not working

nanotuxi:
Extend OpenBot Android app showing the battery state of the serial connected MCU (arduino)


nanotuxi:
Ooops. This was pushed by mistake, as it is only a first graphical test. Sorry for that.

thias15:
No problem, you can also keep this PR open. It will track the changes you make to your branch. Also make sure to add the GUI addition to the same file in the folder layout-land. This will make sure, the element is also there when you rotate the phone.

PR #102, opened 2020-11-10
Add Battery Status for MCU to Android Display

nanotuxi:
1st layout additions for displaying battery status information from MCU on the android display. Not functional yet.

### DIFF
--- a/android/app/src/main/res/layout-land/layout_bottom_sheet.xml
+++ b/android/app/src/main/res/layout-land/layout_bottom_sheet.xml
@@ -107,6 +107,29 @@
         android:orientation="horizontal">
 
         <TextView
+            android:id="@+id/battery_status"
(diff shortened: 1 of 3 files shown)


nanotuxi:
@thias15  Does it make sense to get the values for the battery status from here 
`String data = intent.getStringExtra("data");
                  // Data has the following form: voltage, lWheel, rWheel, obstacle` in class
 `SensorService.java` or would you suggest another approach?

thias15:
Yes, you just have to split the String, then extract the battery voltage value and broadcast it to the UI Thread. There you can update the GUI elements. Are you still working on this? If not, I can also quickly implement this. It's been on the ToDo list for a while...

nanotuxi:
No, it's ready here on my machine, but I simply don't have enough time for the Openbot project anymore. Sorry.

Issue #15, opened 2020-09-02
[Feature] Button to switch camera


Tmin10:
Hello. I run the app on my phone but it display picture from front camera. After debugging I realize that I need to change https://github.com/intel-isl/OpenBot/blob/master/android/app/src/main/java/org/openbot/CameraActivity.java#L110 to 0 (LENS_FACING_FRONT) and it working fine... I don't know why front and back are reversed in my case but it could help you in future for debug the same problems.

My device is Asus Zenfone Max Pro (M2) ZB631KL, Android 9. Other cameras apps working fine.

thias15:
We have not encountered this on any phone before. Thanks for opening this issue, we will keep an eye on it. I will close it for now, since I don't expect this to be a general issue.

Tmin10:
Yep, thanks. Maybe in future it is possible to add camera switch into the settings.

eagleanurag:
> 
> 
> Hello. I run the app on my phone but it display picture from front camera. After debugging I realize that I need to change https://github.com/intel-isl/OpenBot/blob/master/android/app/src/main/java/org/openbot/CameraActivity.java#L110 to 0 (LENS_FACING_FRONT) and it working fine... I don't know why front and back are reversed in my case but it could help you in future for debug the same problems.
> 
> My device is Asus Zenfone Max Pro (M2) ZB631KL, Android 9. Other cameras apps working fine.

tin my case also when i used my old Samsung J7 Max by default front camera opens

vkuehn:
Hi I have a solution for that, can you kindly assign it to me ?
Sorry it will be not fully compliant with your naming rules..this time

vkuehn:
is there an issue why it is a ToDo again ?

PR #115, opened 2020-11-17
Button to switch camera

vkuehn:
Hi, this PR adds a Button to switch between front and back camera as a source. This is related to
[Feature] Button to switch camera #15


thias15:
Thanks a lot. Just tested it. There are some minor issues. 
- the app crashes, when the phone is rotated
- the button is quite large (let's use a toggle switch, like for the other items)
- the crop and frame view do not have to be visible, they are not populated anyways

Can you quickly fix those? Otherwise, I can also do it...

thias15:
So I have updated the above points, but from the browser on my phone. ;) Seems I broke the style. Will fix it tomorrow and merge your PR. Again, thanks a lot!

nanotuxi:
Don't forget to make changes in the layout-land, too. Otherwise the app may not work as expected.

thias15:
@nanotuxi yes. This is already done.

Issue #120, opened 2020-11-23
Web server for log uploading


sanyatuning:
Hi @thias15!
I started working on a simple python HTTP server for log uploading.
I've added zeroconf to the server for auto-discovery if connected to the same network.
I've using Kotlin and NsdManager in the Android app.
What do you think?

thias15:
Sounds good. But our preference would be to keep the app in Java.

sanyatuning:
I've refactored the new classes to java and created a pull request

thias15:
Thank you very much. I'm very busy at the moment, but will review it as soon as possible and merge.

PR #121, opened 2020-11-23
WIP: Web server for log uploading

sanyatuning:
simple python HTTP server for log uploading
zeroconf added to the server for auto-discovery if connected to the same network

see Issue #120

Preview
openbot

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -46,5 +46,6 @@ dependencies {
     implementation 'com.github.felHR85:UsbSerial:6.1.0'
     implementation 'androidx.coordinatorlayout:coordinatorlayout:1.1.0'
     implementation 'org.zeroturnaround:zt-zip:1.14'
+    implementation 'com.loopj.android:android-async-http:1.4.9'
 
 }
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -6,6 +6,7 @@
     <uses-permission android:name="android.permission.CAMERA" />
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
     <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
+    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.INTERNET" />
 
     <uses-feature android:name="android.hardware.camera" android:required="true"/>
@@ -23,7 +24,8 @@
         android:label="@string/app_name"
         android:roundIcon="@drawable/openbot_icon"
         android:supportsRtl="true"
-        android:theme="@style/AppTheme">
+        android:theme="@style/AppTheme"
+        android:usesCleartextTraffic="true">
 
 
         <activity
--- a/android/app/src/main/java/org/openbot/CameraActivity.java
+++ b/android/app/src/main/java/org/openbot/CameraActivity.java
@@ -151,6 +151,8 @@ public abstract class CameraActivity extends AppCompatActivity
   protected String logFolder;
   private boolean loggingEnabled;
   private Intent intentSensorService;
+  private NsdService nsdService;
+  private UploadService uploadService;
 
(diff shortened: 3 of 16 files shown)


thias15:
This looks amazing. Let me know when you would like me to review it to be merged. Also, it would be nice to be able to enable/disable this feature in the app.

sanyatuning:
@thias15 later I would like to create a web interface for retraining models based on new logs.
I'm not a jupyter notebook not an ML guru, but I think it is possible.

thias15:
This sounds like a good idea, and yes it should be doable. It doesn't have to be in Juypter, we just use it to make it easier for people to train models. The python web server, could just trigger a training script that updates an existing policy with new data that comes in. This could even go in the direction of reinforcement learning. The robot will stream its experience constantly to the web server and the web server will send the updated policy to the robot.

thias15:
Finally got around to test it. I will extend the documentation a bit. It was working for me initially but then stopped working. I do see in the Android app that the upload was started at http://127.0.0.1:8000. But it does not seem to finish. After about a minute it end up in `onFailure` in the uploadService. On the server side it shows `GET /favicon.ico`. Anything that I'm missing?

I did run the server with sudo to be sure it has write permissions. I did notice that if I run it with `sudo adev runserver main.py` it starts a dev server at `http://localhost:8000` and an aux server at http://localhost:8001. However, if I run it with `python main.py` it only starts a server at `http://0.0.0.0:8000/`. That all seems fine, even though I do not quite understand what's the difference here. 

Could you explain in a bit more detail, the process to get this to work. Thanks.

thias15:
@sanyatuning I'm not really an expert with http servers. But shouldn't the Android phone connect to an actual IP address on the network rather than localhost? I done some testing today. If I put recording folders manually, they are detected and the preview generation is also working fine. But the upload is not working for me. I'm pretty confident that the problem sits in front of the screen. So I would be grateful, if you could help me figure this out.

sanyatuning:
It should connect to the server as you mentioned.
Please try again, with my new commit.
If not works pls send your logs. It should look like this:
```
$ python main.py 
Dataset dir: /home/sandor/repos/OpenBot/policy/dataset
Skip address 127.0.0.1 @ interface lo
Found address 192.168.0.17 @ interface wlp2s0
Registration of a service, press Ctrl-C to exit...
======== Running on http://0.0.0.0:8000 ========
```

sanyatuning:
@thias15 FYKI
I also added auto-upload old files when the server found (and delete when the upload success)

thias15:
It did connect for me now.
Screen Shot 2020-12-03 at 9 26 48 AM


However, it throws the following error when trying to upload:
```
2020-12-03 09:12:52.265 20839-20894/org.openbot E/AndroidRuntime: FATAL EXCEPTION: inference
    Process: org.openbot, PID: 20839
    java.lang.NullPointerException: Attempt to invoke virtual method 'boolean java.lang.String.isEmpty()' on a null object reference
        at org.openbot.UploadService.upload(UploadService.java:77)
        at org.openbot.CameraActivity.lambda$stopLogging$1$CameraActivity(CameraActivity.java:936)
        at org.openbot.-$$Lambda$CameraActivity$q5CrdoM7lQb1yeKi9jPH5-7ofGw.run(Unknown Source:2)
        at android.os.Handler.handleCallback(Handler.java:900)
        at android.os.Handler.dispatchMessage(Handler.java:103)
        at android.os.Looper.loop(Looper.java:219)
        at android.os.HandlerThread.run(HandlerThread.java:67)
```

That's an easy fix. But after fixing it and restarting the server again, it won't connect. Is it possible that the IP/port gets blocked (not released properly) when I shutdown the server?

thias15:
@sanyatuning thanks for this great contribution. I have updated the README and merged.

sanyatuning:
Thanks for the fixes and the merge, meanwhile I'm working on the web frontend part.
I will create another PR when I'm done.

thias15:
Awesome, looking forward to it.

thias15 (commented):
All tested.

PR #125, opened 2020-12-09
Restructure policy dir, create python package openbot

sanyatuning:
It will be easier to import training files into the server after this refactor.


sanyatuning:
thanks

thias15 (commented):
All tested.

PR #126, opened 2020-12-10
msesma/persist-settings

msesma:
Current app loses user selection between app runs and even on device configuration changes.  This PR will persist some of them and apply during activity onCreateÑ

Data Logger.
USB connection speed.
Drive mode
Camera
Vehicle control
Model
Device
Threads

Other options will not be persisted 

Data Logger switch (logging is takes a lot of space)
Controller switch (For safety reasons)
Connection switch (It changes automatically if a device is connected)

The code consist in a new shared preferences class with methods for each settings and the calls inn the main activity during onCreate for restoring values and on each change listener to store the new values.

I have removed the configuration change code that maintained the camera selected as it is no longer necessary: The camera selection is persisted as any other setting.


thias15 (commented):
Looks great. There is just one minor issue and I don't see in the code where it comes from. When selecting as device GPU and restarting app, it will load NNAPI and vice versa. The CPU selection works as expected. Could you please investigate this issue?

msesma:
Sure, great catch!

thias15:
If you just push to your branch, the PR will be updated automatically.

msesma:
The issue was that the array and the enum didn´t have the same order
`    <string name="device">Device</string>
    <string-array name="devices">
        <item>CPU</item>
        <item>GPU</item>
        <item>NNAPI</item>
    </string-array>`

`  public enum Device {
    CPU,
    NNAPI,
    GPU
  }`

thias15:
Awesome. So you likely fixed a bug that nobody has noticed thus far. ;) Great work!

msesma:
It worked because the enum instance selection was done based on the text selected. And at start the default selection was CPU that was the first item in both the array and the enum. 
The real solution would be filling the spinner with the enum names instead having separate strings arrays... but that is another refactor.

thias15:
@msesma, yes agreed. One minor thing. I saw you refactored setLogMode to setLoggerMode to be consistent with the preference manager. Also the GUI elements carry the Logger name (e.g. loggerSwitchCompat, loggerSpinner). However, I would actually prefer to rename all of those elements from Logger to Log. The reason is that there is also a `LOGGER` which is used for the debug log messages. What do you think?

msesma:
Makes sense. I´ll do it. 
Thanks for pointing it out.

thias15:
Ok, I'll merge after this. Thanks!

msesma:
Done

PR #128, opened 2020-12-13
Dhruv2295/code refactor

dhruv2295:
Migration of dependencies to androidX.
Basic code cleanup to remove redundant initialisers.

### DIFF
--- a/android/app/src/main/java/org/openbot/CameraActivity.java
+++ b/android/app/src/main/java/org/openbot/CameraActivity.java
@@ -104,12 +104,12 @@ public abstract class CameraActivity extends AppCompatActivity
   private int cameraSelection = CameraCharacteristics.LENS_FACING_BACK;
   protected int previewWidth = 0;
   protected int previewHeight = 0;
-  private boolean debug = false;
+  private final boolean debug = false;
   private Handler handler;
   private HandlerThread handlerThread;
   private boolean useCamera2API;
   private boolean isProcessingFrame = false;
-  private byte[][] yuvBytes = new byte[3][];
+  private final byte[][] yuvBytes = new byte[3][];
   private int[] rgbBytes = null;
   private int yRowStride;
   private Runnable postInferenceCallback;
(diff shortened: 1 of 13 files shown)


thias15 (changes requested):
Upgrade to API 30 causes app crash when logging. The permission handling for storage seems to have changed.

android/app/src/main/java/org/openbot/CameraActivity.java
```diff
@@ -240,11 +239,7 @@ protected void onCreate(final Bundle savedInstanceState) {
         new ViewTreeObserver.OnGlobalLayoutListener() {
           @Override
           public void onGlobalLayout() {
-            if (Build.VERSION.SDK_INT < Build.VERSION_CODES.JELLY_BEAN) {
```
thias15, line 243: What about old Android versions?

android/app/src/main/java/org/openbot/NetworkActivity.java
```diff
@@ -193,81 +193,78 @@ protected void processImage() {
...
-                      // showFrameInfo(previewWidth + "x" + previewHeight);
-                      // showCropInfo(croppedBitmap.getWidth() + "x" + croppedBitmap.getHeight());
-                      showInference(lastProcessingTimeMs + "ms");
-                    }
-                  });
-            }
+            updateVehicleState();
+
+            runOnUiThread(
+                new Runnable() {
```
thias15, line 260: Why use lambda previously but not here?

android/app/src/main/java/org/openbot/env/UsbConnection.java
```diff
@@ -53,29 +52,22 @@ public UsbConnection(Context context, int baudRate) {
...
-                    public void run() {
-                      onSerialDataReceived(dataStr);
-                    }
-                  });
-            }
-          } catch (UnsupportedEncodingException e) {
-            LOGGER.e("Error receiving USB data");
-          }
+  private final UsbSerialInterface.UsbReadCallback callback =
+      data -> {
```
thias15, line 56: What happens on error, since try/catch is removed?

android/app/src/main/java/org/openbot/CameraActivity.java
dhruv2295: App's minimum supported version is 21. This check was checking for 19.

android/app/src/main/java/org/openbot/NetworkActivity.java
dhruv2295: There's no global converter in android studio to convert all these anonymous functions to lambdas at once. 
So it's more of an iterative process, where I keep replacing them as and when I find them.

android/app/src/main/java/org/openbot/env/UsbConnection.java
dhruv2295: It was an android studio recommendation that this catch block would never be called. I'll add it back. Better safer.

android/app/src/main/java/org/openbot/NetworkActivity.java
dhruv2295: I've converted this and pushed another commit

android/app/src/main/java/org/openbot/CameraActivity.java
thias15: Yes, makes sense. Android 5.0 (API level 21) = Lollipop.

PR #133, opened 2020-12-27
Ivo/sending data to controller

izivkov:


izivkov:
This PR enables the bot app to work with OpenBot's phone-based controller. It now supports 2-way communications, so the controller can update its UI based on bot's settings.

thias15 (approved):

PR #135, opened 2020-12-31
download networks

thias15:
- gradle script for downloading the network tflite files during before build


thias15 (commented):
Github build action still works.

PR #136, opened 2020-12-31
Firmware trim motors

vkuehn:
vendors of cheap dc motor often produce motors with slightly different performance. this changes allows to enable dynamic trimming. When active the performance is monitored on straight runs to mitigate the issue

### DIFF
--- a/firmware/openbot_v1_nano/openbot_v1_nano.ino
+++ b/firmware/openbot_v1_nano/openbot_v1_nano.ino
@@ -40,19 +40,22 @@
 //------------------------------------------------------//
 
 // Setup the OpenBot version (DIY,PCB_V1,PCB_V2)
-#define OPENBOT DIY
+#define OPENBOT PCB_V1
 
 // Enable/Disable voltage divider (1,0)
-#define HAS_VOLTAGE_DIVIDER 0
+#define HAS_VOLTAGE_DIVIDER 1
 
 // Enable/Disable indicators (1,0)
 #define HAS_INDICATORS 0
 
 // Enable/Disable speed sensors (1,0)
-#define HAS_SPEED_SENSORS 0
+#define HAS_SPEED_SENSORS 1
+
+// Enable/Disable dynamic motor trimming.Only works with Speed Sensors (1,0)
+#define USE_SPEED_TRIM 0
 
 // Enable/Disable sonar (1,0)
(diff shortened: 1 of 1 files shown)


thias15:
Here are some thoughts
- The trimming also makes sense when going in reverse. So any time ctrl_left==ctrl_right
- To me ctrl_left_trimmed and ctrl_right_trimmed as variable names makes more sense
- Probably not necessary to send back the trimmed values or even the trim value. The Android app doesn't really care about the specifics of the car, it just expects it to do the right thing.

thias15:
In addition:
- Currently there can be overflow. If ctrl_left and ctrl_right are already at the max (255), trim/2 is added to one side. This result in an invalid ctrl signal. It is probably better to subtract the absolute trim value from one side if the controls are positive and add it to one side if they are negative. 
- Right now it will only trim when going straight. How about comparing the ratio between ctrl_left/ctrl_right and speed_left/speed_right. Then trimming can always be active

vkuehn:
when driving in circles trimming probably is to difficult there it is only active when driving straight. I have to rethink about the overflow. it didn't happen to me but you could be right. I'll probably update the code for it

vkuehn:
pls review

thias15:
@vkuehn Sorry for the delay. I did test it and encountered some issues. For me it did not compensate properly when going straight. And when going backwards, the car behaves very strangely for some reason. How well did the new firmware work in your tests?

vkuehn:
Strange habits was what I saw with the firmware that is why I started that change.

I have to admit I still did struggle running the OpenBot around as of a lack of a properly connected controller. 
Through an USB hub and a cable I did use a ps3 controller. PS4 over Bluetooth but I had to give that one back..
Mainly I did test running forward.

Compensation now is just finding a middle value between the faster and the slower wheel. It is not a complex normalization as it is running on the Arduino.

@thias15 I'll think again how to make that one better. Still difficult to understand why there are sometimes these strange irritations.

vkuehn:
Sorry took me ages to find a solution to test the firmware without a mobile phone as there is no web interface with direct debugging in OpenBot right now. Not sure why python fails to talk reliable to that firmware so did it with node red also. However it turns out that all odd behavior is coming from small problems with the hardware. Like not properly soldered cables or bad tire position on these cheap motors

PR #144, opened 2021-01-10
Download models from server

sanyatuning:
The android app will download "published" tflite files from the server if the server discovered on the local network.

### DIFF
--- a/android/app/src/main/java/org/openbot/env/SharedPreferencesManager.java
+++ b/android/app/src/main/java/org/openbot/env/SharedPreferencesManager.java
@@ -3,6 +3,7 @@ package org.openbot.env;
 import android.content.Context;
 import android.content.SharedPreferences;
 import org.openbot.common.Enums;
+import org.openbot.tflite.Model;
 import org.openbot.tflite.Network;
 
 public class SharedPreferencesManager {
@@ -18,8 +19,8 @@ public class SharedPreferencesManager {
   private static final String SPEED_MODE = "SPEED_MODE";
   private static final int DEFAULT_DRIVE_MODE = Enums.DriveMode.GAME.ordinal();
   private static final String DRIVE_MODE = "DRIVE_MODE";
-  private static final int DEFAULT_MODEL = Network.Model.DETECTOR_V1_1_0_Q.ordinal();
-  private static final String MODEL = "MODEL";
+  private static final String DEFAULT_MODEL = Model.DETECTOR_V1_1_0_Q.toString();
+  private static final String MODEL = "MODEL_NAME";
   private static final int DEFAULT_DEVICE = Network.Device.CPU.ordinal();
   private static final String DEVICE = "DEVICE";
   private static final int DEFAULT_NUM_THREAD = 1;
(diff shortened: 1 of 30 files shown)


thias15:
Screen Shot 2021-01-12 at 7 19 03 PM
For me the model gets continuously downloaded by the phone after it was added to the published models. See the attached log, also on the Android phone the toast is displayed every few seconds. I think a published model should only be downloaded once (e.g. onClick publish model).

Also, the long model name can cause some issue in the app layout.

sanyatuning:
I don't quite understand why is this, there is a file time check before download.
I've added some logs with file times to the app, pls test, and provide android logs if not works.

thias15:
@sanyatuning it still does not work in the sense that it keeps downloading models. Do you also observe this behaviour on your end?

There is a NSD warning:
```
2021-01-13 17:12:33.307 26842-27071/? W/NSD: java.lang.IllegalArgumentException: listener already in use
        at com.android.internal.util.Preconditions.checkArgument(Preconditions.java:50)
        at android.net.nsd.NsdManager.putListener(NsdManager.java:448)
        at android.net.nsd.NsdManager.resolveService(NsdManager.java:616)
        at org.openbot.robot.NsdService$1.onServiceFound(NsdService.java:30)
        at android.net.nsd.NsdManager$ServiceHandler.handleMessage(NsdManager.java:387)
        at android.os.Handler.dispatchMessage(Handler.java:110)
        at android.os.Looper.loop(Looper.java:219)
        at android.os.HandlerThread.run(HandlerThread.java:67)
```

But other than this I'm not seeing any debug messages with the timestamps that you added. Very strange!

It might be nice to add an option to `unpublish` models, so they will no longer be served for download and maybe even deleted from the phone.

thias15:
On the phone I was testing with the clock was not set to automatic. The time was about 5 minutes wrong. It seems that was the issue.

sanyatuning:
I saw the NSD error on my side and added a catch block for that.
I can set the file modification time to what the server sent. 
I think this will solve this, what do you think?

thias15 (approved):

thias15:
Yes, I think using server time would be best. For now I have added a note to the readme file.

thias15:
Also, I think there should be an option to delete models, either from server or in app. Because if people test many models, it might get very confusing.

sanyatuning:
Good point I will work on it.
Do you think auto delete when "unpublish" would be OK?

thias15:
> Do you think auto delete when "unpublish" would be OK?

Yes, I think that's fine. We are working on a refactor of the app anyways. 
As part of that we can also add controls to the app to select models from phone storage, delete models, etc.

thias15:
@sanyatuning just wanted to follow up on the status of the `unpublish` feature. I'm planning to make the next release in the next few days. Would be great to merge this PR before.

sanyatuning:
@thias15 I've added some changes, but not tested the android part.

thias15:
@sanyatuning when I publish a model from the frontend, it does not show up under published models. Ergo, I also do n9ot get the `unpublish` option. However, I do see it in the folder on the laptop and it is also transferred to the phone successfully.

PR #154, opened 2021-01-27
Dhruv2295/vehicle communication

dhruv2295:
All new vehicle communication screen  with interactive UI


thias15 (commented):
There are two little issues. 
(1) When exiting the robot screen, USB is disconnected (toast message). When going into the Free Roam screen, USB still works, I assume you automatically connect. However, if going back to the main screen and then back into Free Roam it no longer works. If I look into the settings menu, USB shows as connect but device name shows "No Device". So there seems to be some issue with USB connection handling. In the Free Roam mode, the user has no idea about the USB connection. I think we need to show the status of the USB connection, and if it is disconnected, user should be able to click on the USB icon to get to the settings menu in order to connect.

(2) When the phone is rotated in the Free Roam GUI, the app changes to the Robot GUI.

One more general comment unrelated to this PR. We should check the GameController connection status. Mine was out of battery and disconnected while I was driving. The car kept driving as a result. On disconnect of the GameController, we should send (0,0) to the car.

dhruv2295:
1. USB status icon added to free roam screen, and connection control moved out to global main activity.
2. screen changing to robot GUI on rotate fixed. 

based on the code, phone controller is sending 0,0 to car on being disconnected. perhaps out of battery device is an edge case which maybe doesn't send disconnect signal like it should. i'll analyse it more.

thias15:
The UI elements do not update with the USB data (battery, sonar, rpm readings). Only the control signal, speedometer, steering, update correctly. If I go into the screen without USB connect, the USB connection shows as connected for some reason.

thias15 (commented):
Tested. LiveData for vehicle interface is buggy and should be fixed with next PR.

android/app/src/main/java/org/openbot/env/Vehicle.java
```diff
@@ -156,34 +167,42 @@ public void connectUsb() {
...
-                "c%d,%d\n",
-                (int) (getControl().getLeft()),
-                (int) (getControl().getRight())));
-
-    Objects.requireNonNull(usbConnection).stopUsbConnection();
-    usbConnection = null;
-    usbConnected = false;
+    if (usbConnection != null) {
+      usbConnection.send(
+          String.format(Locale.US, "c%d,%d\n", (int) (getLeftSpeed()), (int) (getRightSpeed())));
```
aimiebell, line 173: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [FS: Format string should use %n rather than n (VA_FORMAT_STRING_USES_NEWLINE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#fs-format-string-should-use-n-rather-than-n-va-format-string-uses-newline).
This format string includes a newline character (\n). In format strings, it is generally  preferable to use %n, which will produce the platform-specific line separator.

PR #155, opened 2021-02-09
Ivo/add wifi p2p connection

izivkov:
What is in this merge request:
- Added another transport layer interface WiFiDirectConnection. Can be used in place of NearbyConnection, but not used at the moment.
- Fixed a couple of display issues in the controller. For example, the `Drive Mode` button in the initial screen had initial value of `0` instead of `Dual Drive`
- Updated the technical controller document to fix some errors in the commands, as well as ass the new connection type.


thias15:
@izivkov. I saw that the WiFi Direct library does not seem to be maintained, last update 4 years ago. Is there an alternative? Did you test it in comparison with Nearby connection and find any benefit? I guess the main benefit would be video streaming capability?

izivkov:
Ok, I looked into various libraries, and would not find anything that is
more current. Having said that, it us just a wrapper for Google's wifi p2p.
Using the Google's code directly makes it complex and difficult to
maintain.

However, I see your point. Let me remove the wifi part and I will create
another MR. It is not 100% reliable anyway. Maybe the right way to do this
is to have our own layer on top of the Google's. This way we have more
control.

Go ahead and cancel this MR, and I will have another one just with the
fixes.

Thanks,
Ivo

PR #157, opened 2021-02-11
Ivo/fixes after main code refacturing

izivkov:
Removed WiFi p2p and made some code fixes to be able to merge with master.


android/app/src/main/java/org/openbot/common/Utils.java
```diff
@@ -22,7 +22,7 @@ public static JSONObject getStatus(
       boolean loggingEnabled,
       boolean noiseEnabled,
       boolean networkEnabled,
-      int driveMode,
+      String driveMode,
```
thias15, line 25: Why does this need to be a string? This change breaks other parts of the app, where the driveMode.getValue() is passed.

thias15: This can easily be fixed by replacing it with driveMode.toString(). But the question is what the rationale behind this change is.

android/app/src/main/java/org/openbot/env/NearbyConnection.java
```diff
@@ -165,21 +191,21 @@ private void stopDiscovery() {
...
-    NearbyConnection.context = context;
-    NearbyConnection.payloadCallback = payloadCallback;
+  @Override
+  public void connect(Context context) {
+    this.context = context;
     connectionsClient = Nearby.getConnectionsClient(context);
 
     // make sure we are not connecting
-    disconnect();
+    disconnect(context);
```
thias15, line 200: Previously this did not require passing the context. Why is the context needed now?

android/app/src/main/java/org/openbot/common/Utils.java
izivkov: The reason it is a string is that we would not want to controller to interpret these numeric values. For example, if we pass 1, this has meaning in the bot app, but the controller does not need to know that 1 means "DUAL-DRIVE" for example. The controller would ideally just display what it is sent. This decouples them, in case these values change in the future, or add more options.

thias15:
I have pushed some minor fixes and tested.

android/app/src/main/java/org/openbot/common/Utils.java
thias15: Ok, makes sense. I have already updated this in the new UI fragment.

PR #158, opened 2021-02-13
AI screen

dhruv2295:
Model Screen for robot to run itself


thias15:
Currently none of the AI models are working. It seems the models are not executed since the inference time is shown as 0ms.

thias15:
When device is changed to GPU and then back to CPU, the number of threads is -1.

PR #162, opened 2021-02-26
Ivo/video streaming 2

izivkov:
- Added RTSP server for FreeRoamFragment.
- Added NetworkServiceConnection to connecting on LAN. Still some issues with reliability when connecting and disconnecting multiple time, so we are still using NearbyConnection
- Added a custom Video component to controller based on VLC library to display streaming video.


thias15 (commented):
The main issue that still needs to be addressed is the latency. It is very high right now.

izivkov:
- The latest push resolves issues with commands not making it from the controller to the bot. This helped with stability of code.
- Next I will address performance of video. Not sure how much this can be improved with the current library. 
- When Marcus is ready with WebRTC server, the current RTSP server can be replaced, by implementing the IVideoServer interface.
- If we can improve the video latency, the rest of the code is pretty stable at this point, but I don't think the code is ready to merged. It is here to collaborate and test with other developers for now.

dhruv2295:
it seems something is up with phone controller. connection between 2 devices doesn't seem to establish.

izivkov:
@dhruv2295 Just to cover the basics, both the app and the controller are on the same WiFi network, right? 
I will retest, but I did some sanity tests before I pushed to code.

dhruv2295:
That's right. Same network

izivkov:
@dhruv2295 Is this the first time you try with the NetworkServiceConnection? 
I don't see this, so can you please help debug the issue? Maybe the app cannot find the controller service. 
Can you see if the onServiceFound() method is called in in the NetworkServiceConnection.java on the bot side? If so, can you next see if onServiceResolved() method is called? If so, the connection should be established.
Thanks in advance.

izivkov:
@dhruv2295 Another silly questions, but did you use the old bot software with the this controller? We switched the convection type from NearbBy to NetworkServiceConnection, so the this new controller will not connect to the old main app. So we need to use both Controller and Bot software from this MR together.

dhruv2295:
@thias15 did I use old or new boot software? 😅

dhruv2295:
@izivkov you mean firmware flashing? Or the mobile application?

izivkov:
I pushed some code to the PR again. Here are some points:
- Fixed a connection issue between the controller and bot when both apps are running on the same device. The controller tries to connect the app in the same device and not on the remote app.
- I tried various ways replace the RTSP server library. There is another library here: https://github.com/pedroSG94/rtmp-rtsp-stream-client-java, but I find it unreliable. So I think the best option is still libsteaming (https://github.com/fyhertz/libstreaming). Maybe we can replace it in the future with another library or another protocol like WebRTC.
- Another issue is using `com.github.ar-android:libstreaming:1.0.0` git repository instead of using libstraming directly. The problem is that `libstreaming` has not be updated to the `jitpack.io` directory since version 3-beta, which. is very old. The only other way is to include this library in source form as a library module, which adds maintenance issues. I'm looking for suggestions on this.
- In summary: 
    1. This version is stable as far as connection, sending commands, etc.
    2. The latency for the video is still large (2 secs)
- Since this PR improves the stability of the app, I think it is worth merging in. We might choose to remove the video functionality if we deem necessary .

thias15:
The library by predo seems to better in the sense that it is much more actively maintained. In what sense is it unreliable. How does the latency compare?

thias15:
I have just resolved the conflicts. Let me do some further testing before merging. But with the current latency the video streaming is not really usable. So not sure if it's better to disable it. In any case it's probably not a bad idea to have a switch for turning it on/off.

android/app/src/main/java/org/openbot/robot/CameraActivity.java
```diff
@@ -937,9 +954,7 @@ private void setControlMode(ControlMode controlMode) {
   }
 
   private void connectPhoneController() {
-    if (!phoneController.isConnected()) {
```
thias15, line 940: I think it might make sense to keep this check. Or is this no longer needed?

android/app/src/main/java/org/openbot/robot/FreeRoamFragment.java
```diff
@@ -43,6 +42,8 @@ public View onCreateView(
   public void onViewCreated(@NonNull View view, @Nullable Bundle savedInstanceState) {
     super.onViewCreated(view, savedInstanceState);
 
+    phoneController.setView(binding.videoWondow);
```
thias15, line 45: Typo --> videoWindow

android/app/src/main/res/layout-land/fragment_free_roam.xml
```diff
@@ -6,7 +6,18 @@
     android:layout_width="match_parent"
     android:layout_height="match_parent"
     tools:context=".robot.FreeRoamFragment">
-    
+
+    <org.openbot.customview.AutoFitSurfaceView
```
thias15, line 10: So I guess the video streaming will only work in the free roam fragment.

android/build.gradle
```diff
@@ -8,7 +8,7 @@ buildscript {
         mavenLocal()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:4.1.1'
+        classpath 'com.android.tools.build:gradle:4.2.0-beta04'
```
thias15, line 11: I think it's best to stay with stable versions and not beta unless it is necessary.

android/app/src/main/res/layout/fragment_free_roam.xml
```diff
@@ -7,6 +7,17 @@
     android:layout_height="match_parent"
     tools:context=".robot.FreeRoamFragment">
 
+    <org.openbot.customview.AutoFitSurfaceView
+        android:id="@+id/video_wondow"
```
thias15, line 11: See comments above.

android/app/build.gradle
```diff
@@ -8,7 +8,7 @@ android {
     defaultConfig {
         applicationId "org.openbot"
         minSdkVersion 21
-        targetSdkVersion 29
+        targetSdkVersion 30
```
thias15, line 11: This needs to be kept at 29, otherwise the logging does not work. There has been a change in storage permissions in API30.

android/.gitignore
```diff
@@ -3,6 +3,8 @@
 *.ap_
 *.aab
 
+notes.txt
```
thias15, line 6: change to *.txt

thias15 (commented):
Please review the comments.

thias15:
Also before pushing I would recommend to run ./gradlew applyStyle inside the android folder. This will apply the style to all java files and will allow to pass the style check.

android/app/src/main/res/layout-land/fragment_free_roam.xml
izivkov: Unfortunately, this is correct. The `CameraConnectionFragment` uses `TextureView` (as opposed t `SurficeView`, which is not compatible with RTSP library. Changing to SurfaceView involves a lit of changes to the main code, which I'm not very comfortable making, especially since we are planning to replace RTSP with WebRTC. We can also experiment with other RTSP libraries in the future. As I said, I tried [this](https://github.com/pedroSG94/RTSP-Server) library with intermittent success, and still have the code, if somebody is will ing to get it to work.

android/app/src/main/java/org/openbot/robot/CameraActivity.java
izivkov: I moved the knowledge of whether the controller is connected or not to the controller itself for two reasons:
1. Simplify the client.
2. We now do `soft` connect and `disconnect`, which means we connection is still open to minimize connection/disconnection, but it still appears to the client like the they are connected/disconnected. This way the client code does not change, while eliminating problems with frequent disconnections. It takes some time for the devices to discover and connect to each other. If somebody switches quickly between controllers, this will interrupt the discovery process and will cause  instability.

android/app/src/main/java/org/openbot/robot/CameraActivity.java
```diff
@@ -922,7 +939,7 @@ private void setControlMode(ControlMode controlMode) {
           disconnectPhoneController();
           break;
         case PHONE:
-          handleControllerEvents();
```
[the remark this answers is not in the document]
izivkov: Sorry, my bad. I was trying to resolve with multiple subscriptions, so I moved this call to the `onCreate` method in the start  of the app. Multiple subscription are now eliminated in the `ControllerToBotEventBus`, so this is not an issue, and I have added this back.

izivkov:
> I have just resolved the conflicts. Let me do some further testing before merging. But with the current latency the video streaming is not really usable. So not sure if it's better to disable it. In any case it's probably not a bad idea to have a switch for turning it on/off.

I will disable the video. Thanks.

android/app/src/main/java/org/openbot/robot/CameraActivity.java
thias15: Great.

thias15:
Thanks. I will test it again.

android/app/src/main/java/org/openbot/env/PhoneController.java
```diff
@@ -27,13 +52,22 @@ public void dataReceived(String commandStr) {
...
-    connection.connect(context);
+    if (!connection.isConnected()) {
+      connection.init(context);
+      connection.connect(context);
+    } else {
+      connection.start();
+    }
   }
 
   public void disconnect(Context context) {
```
thias15, line 69: Seems context is no longer needed.

android/app/src/main/java/org/openbot/robot/CameraActivity.java
```diff
@@ -682,6 +683,20 @@ public void onRequestPermissionsResult(
...
+      case REQUEST_RECORD_PERMISSION_CONTROLLER:
+        if (grantResults.length > 0 && grantResults[0] == PackageManager.PERMISSION_GRANTED) {
+          // ask for recording permissions here
+
+          ActivityCompat.requestPermissions(
+              this, new String[] {RECORD_AUDIO}, REQUEST_RECORD_PERMISSION_CONTROLLER);
+
+        } else {
+          if (ActivityCompat.shouldShowRequestPermissionRationale(this, PERMISSION_LOCATION)) {
+            Toast.makeText(this, R.string.location_permission_denied_controller, Toast.LENGTH_LONG)
```
thias15, line 695: There should be a new string defined for the audio permission, not display the location permission text.

android/app/src/main/res/layout/fragment_free_roam.xml
```diff
@@ -7,6 +7,17 @@
     android:layout_height="match_parent"
     tools:context=".robot.FreeRoamFragment">
 
+    <org.openbot.customview.AutoFitSurfaceView
```
thias15, line 10: This invisible texture is not ideal, but we can keep it for now.

PR #164, opened 2021-03-14
Global Initialisation of Vehicle Class

dhruv2295:
1. vehicle initialisation in application class.
2. prevent usb disconnection on orientation change
3. move classes to relevant directories


thias15 (commented):
Looks good. Need to test it still.

thias15:
Tested and works well. 
Please:
- resolve conflicts
- fix usb toggle in settings menu (does not update if usb was disconnected in default activity) --> maybe just try to auto-connect

PR #169, opened 2021-03-18
fix thread closure on screent timeout

dhruv2295:
prevent model closure on screen timeout

### DIFF
--- a/android/app/src/main/java/org/openbot/ai/AIFragment.java
+++ b/android/app/src/main/java/org/openbot/ai/AIFragment.java
@@ -145,8 +145,7 @@ public class AIFragment extends CameraFragment implements ServerListener {
           setNumThreads(--numThreads);
           binding.threads.setText(String.valueOf(numThreads));
         });
-    BottomSheetBehavior.from(binding.loggerBottomSheet)
-        .setState(BottomSheetBehavior.STATE_EXPANDED);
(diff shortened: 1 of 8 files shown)


zlite:
Thanks for this one. It was a super frustrating glitch that got me every time

thias15:
@zlite were you already using the ai fragment or did you also experience something similar on another screen?

zlite:
I've consistently experienced driving freezing after about a minute during training. I had assumed it was due to a screen timeout, but perhaps it was due to something else?

thias15:
Do you mean when driving and collecting training data or when running a policy? If while running a policy, autopilot or detector?

zlite:
While driving and collecting training data. It invariably freezes after about a minute, so I've just learned to train for shorter than that and save before it freezes

thias15:
Yes, that's annoying. What phone are you using? Does the app freeze after that time or just the screen turn off?

thias15:
It sounds to me like it might be another issue. It would be good if you can open an issue describing the problem in as much detail as possible. Then we can investigate further.

thias15:
I kept the screen on which fixed the timeout issue. For the detector issue, I noticed that the MaxAnalyseImageSize is 720 x 720. In the Default activity it's 1280 x 720. I'm. pretty sure that's the core of the problem.

thias15:
Have a look at the `chooseOptimalSize` function in the CameraConnectionFragment. I think we need something like that for the new camera fragment. We cannot just take whatever res the OS gives us.

thias15 (approved):

PR #173, opened 2021-04-02
Ivo/keybaord controller

izivkov:
Added a python program to act as a controller. This program can run on a Linux-based computer (including Raspberry PI) and allow the user to control the robot from the keyboard. Can be used for fast prototyping of new commands, or as a controller from wireless keyboard.

Has not been tested on Windows, but should run as well.


thias15:
This is great. Just one comment, I feel like the control is not very natural. If the keyboard button is released, I would expect throttle to be lowered until 0. Similar with the steering.

izivkov:
> This is great. Just one comment, I feel like the control is not very natural. If the keyboard button is released, I would expect throttle to be lowered until 0. Similar with the steering.

Ok, this is good idea. I will work on this. I will have an update soon, so maybe wait for that.

thias15:
Ok, I will wait and then merge. Another question. Is there any difference between the connection to the Python script and the phone controller now? I'm wondering if we should update the UI to have one more controller option or maybe swap the phone icon for something more generic.

izivkov:
> Ok, I will wait and then merge. Another question. Is there any difference between the connection to the Python script and the phone controller now? I'm wondering if we should update the UI to have one more controller option or maybe swap the phone icon for something more generic.

The python script uses the exact same connection as the android controller does. It locates services on the LAN by service name and type. Both controller and robot app have to be on the same LAN. So, we can call the controller something like WiFi controller, or something like this. But I am not sure the python script will be used a lot outside of prototyping. Maybe for Raspberry PI  the controller could me more useful. 

There is also something called [pygame](https://www.pygame.org/docs/), which allows you to create UI in python for games, so this maybe useful

izivkov:
> This is great. Just one comment, I feel like the control is not very natural. If the keyboard button is released, I would expect throttle to be lowered until 0. Similar with the steering.

One concern I have is how to you make the bot run at a constant speed? If we slow down as soon as we stop pressing the up arrow, the robot will slow down and stop. With the dual drive controls, we can leave them the bot will move at the same speed, but if we let go the keyboard, the robot will stop. On the other hand, if we keep pressing the keyboard, the robot will accelerate.

I can see if we stop pressing the left/right arrows, the robot will straighten, so this makes sence.

thias15:
But even for the phone controller we discussed that it would be nice for the sliders to bounce back if the user releases them. Same logic, it is not very natural to manually move the sliders back down. I think to keep the button/slider pressed to keep a constant speed is expected. It is the same way the bot is controller with the BT controller and pretty much in any video game. For keeping constant speed, we could just have a cruise control key. If you press it, we will keep the same speed until it is pressed again.

thias15:
I think using pygame or even just vlc player to get the video stream on the laptop would be very nice.

izivkov:
Now the controls on the keyboard-controller after the user stops pressing the arrow keys revert back to 0/0 in steps within 4 seconds.

thias15:
Great, I will test it again. I think one second would be enough to stop the car. Four seconds seems quite long if someone needs to stop the car. What do you think? I also noticed that you also removed the flashlight control from the controller app. Did you not get this to work?

thias15:
Hi @izivkov.  I have merged the PR now. Again, thank you very much for your contribution. I feel that the control is still not very natural. In most video games, you would expect that the car will go a full speed when pressing the forward button and stop (coast) once you release the button. Similarly for turning left and right, but if the forward button is still pressed the car would be expected to continue driving forward instead of controls going to 0. I wonder if it would be easier to implement with callbacks on key_pressed and key_release. I came across this [library](https://github.com/boppreh/keyboard#keyboard.KeyboardEvent). What do you think?

Another thing we could do is just provide throttle and steering to the bot and use the functions in the GameController to convert to the left and right control values. But I guess the challenge with the keyboard is that everything is binary, either pressed or release.

izivkov:
Hey @thias15, I was just about to push changes with 1 second revert to zero
instead of 4 seconds.

I am a bit confused. I thought this is what the new controller did, as you
described. If you let go of the forward button (or all buttons), the car
will come to a stop, initially in 4 seconds, but now in one seconds. The
left and right controls will revert back to zero within this time period. I
have not actually tried it on the robot, but I can see on the robot phone,
the values of the controls become zero. Basically, it is not a problem to
implement anything the way we like, without any additional keyboard
library, but just want to make sure I understand what needs to be done.

I am working on a UI version of this program, using PyGame, and will
support video. I still like to keep this version for terminal only
applications, like using ssh to connect to the device running the app. It
should even work with Raspberry Pi Zero.

Ivo

izivkov:
I see you have merged the code, so my latest change did not make it in, but
no problem, will get it next time.

I was just wondering if you see what I see. If you stop pressing the
keyboard, do the controls revert to zero after 4 seconds?

Ivo

PR #182, opened 2021-04-22
Ivo/keyboard controller UI

izivkov:
There are a number of improvements in this PR. The main are:
1. Video has smaller latency, higher resolution and full screen. Also added sound.
2. Network connection type is selected based on availability of WiFi. Now the controller can operate outside using NeabBy connection if no WiFi is present.
3. The drive controls are reverted back to home position after one second of inactivity.
4.  RTSP video stream stopped/started when controller is disconnected/connected. This was a bug in the previous version. 
5. Added Python keyboard controller with video from robot. More work is to be done on this to display some keys, usage, speed, etc.

TODO:
1. The video still does not stream from the "Legacy" screen. Although it is now technically possible to do this, since the new RTSP video supports `TextureView`, there are a number of problems:
 - The video on the robot phone becomes distorted.
 - The existing surface is not GL type, like `OpenGlView`, so it cannot process the information fast enough and the video stream stops on some slower phones.
 2. Add WiFi direct connection. This maybe preferable to current dual connection solution to improve reliability when switching between connection types. On the other hand, the Python scripts generally will not be able to run with WiFi direct, because they run on PC's.


android/app/src/main/java/org/openbot/common/ControlsFragment.java
```diff
@@ -48,7 +48,6 @@ public void onViewCreated(@NonNull View view, @Nullable Bundle savedInstanceStat
         .getWindow()
         .addFlags(android.view.WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON);
 
-    phoneController.init(requireContext());
```
thias15, line 51: Why is this line removed? I believe it is causing the app to crash since the phoneController is not initialized.

izivkov: I wanted to eliminate the need to call `init()` because people forget to call it. Now we are forced to pass the context via the `getInstance(contect)` method.

This latest push should address the connectivity issues. The main problem was that I was returning a new connection from the `ConnectionManager` class when the user called `getConnection()`

Also updated the the latest version of the RTSP server.
In some rear cases, the video stream will fail to start. I will see if there is a fix, but in the mean time, just close the Free Roam fragment by going back, end enter it again.

izivkov:
@thias15 I think your idea to select Nearby if the user does not give camera permissions is good, but there is a problem. The controller does not know that it need to use Nearby, and will try Network connection. The only time they can use nearby if both do not have WiFi connection. I thought of trying Nearby first, but this complicates things and will not work with the python script. So, I keep it simple:
- If the app or the controller see WiFi connection, use network connection.
- Otherwise use nearby.
Note that this creates a problem if one of them has WiFi and the other not. They will try to connect via a different methods, and this will fail.
I also tried to see if the Network connection will fail quickly if it does not see the other device, but the timeout is long or the device will try forever.

thias15:
Hi @izivkov. Yes I agree. My first idea would have been to always start with Nearby and then upgrade to Network if video and WiFi is available. But yes, this would not work for the Python script. My seconds thought would be to add an option to the controller app where the user can select the connection type. But I guess, your current approach has a similar effect. If the user wants to connect via Nearby, they just need to turn off Wifi on both devices. If it is possible to control the Network timeout, this could be another good option to consider in the future. First try network for 10 seconds or so, if it fails, try Nearby. 

I will test the changes in the PR again. Is there anything else, you still want to do as part of this PR or is it ready to merge from your pov? I guess one thing we discussed and I did not see in the changes was to reset the sliders on TouchRelease rather than a timer.  But if this is a bigger endeavor, let's do it in another PR.

thias15:
In testing, I encountered the following issue: on my phone the resolution is not supported by the camera. Seems, like we need to select the resolution adaptively depending on the phone model. We can use the same approach as in the logger fragment or a function like `chooseOptimalSize` in the CameraConnectionFragment.java`.
```
2021-04-26 10:46:12.434 12555-12555/org.openbot E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.openbot, PID: 12555
    com.pedro.encoder.input.video.CameraOpenException: This camera resolution cant be opened
        at com.pedro.encoder.input.video.Camera1ApiManager.start(Camera1ApiManager.java:136)
        at com.pedro.encoder.input.video.Camera1ApiManager.start(Camera1ApiManager.java:124)
        at com.pedro.encoder.input.video.Camera1ApiManager.start(Camera1ApiManager.java:131)
        at com.pedro.rtplibrary.base.Camera1Base.startEncoders(Camera1Base.java:562)
        at com.pedro.rtplibrary.base.Camera1Base.startStream(Camera1Base.java:546)
        at org.openbot.env.RtspServer.startServer(RtspServer.java:144)
        at org.openbot.env.RtspServer.startServer(RtspServer.java:112)
        at org.openbot.env.RtspServer.lambda$init$0$RtspServer(RtspServer.java:53)
        at org.openbot.env.-$$Lambda$RtspServer$kmKwnVsBgIqQhLivjgc2W6-rGQE.runIt(Unknown Source:2)
        at org.openbot.utils.AndGate.update(AndGate.java:37)
        at org.openbot.env.RtspServer.setConnected(RtspServer.java:120)
        at org.openbot.env.PhoneController.connect(PhoneController.java:86)
        at org.openbot.robot.FreeRoamFragment.connectPhoneController(FreeRoamFragment.java:249)
        at org.openbot.robot.FreeRoamFragment.setControlMode(FreeRoamFragment.java:219)
        at org.openbot.robot.FreeRoamFragment.lambda$onViewCreated$0$FreeRoamFragment(FreeRoamFragment.java:60)
        at org.openbot.robot.-$$Lambda$FreeRoamFragment$KXqOJ5FhyUlRwLuVjqYgusRElcE.onClick(Unknown Source:2)
        at android.view.View.performClick(View.java:7359)
        at android.view.View.performClickInternal(View.java:7325)
        at android.view.View.access$3200(View.java:846)
        at android.view.View$PerformClick.run(View.java:27807)
        at android.os.Handler.handleCallback(Handler.java:873)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:214)
        at android.app.ActivityThread.main(ActivityThread.java:7073)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:494)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:965)
```

thias15:
Just pushed a fix for this error. Let me know if there is anything else you would like to add to this PR, otherwise I'll go ahead and merge. I still find the connection not to be 100% stable, sometimes freezing or not connecting. But we can look into this in the future.

thias15:
After some more testing, I found that python now reports `OpenCV: Couldn't read video stream from ...` on some phones (e.g. S20FE).
It was working before the recent changes and still works on the another phone. Strange.

thias15:
So, after multiple tries in eventually works. I think we need to do some more testing and make the video streaming more robust.

izivkov:
- OK, so the new resolution of 640x360 caused some problems. I saw your fix for the android controller. Does it work with all phones now? I'f you don't mind, in the future, I will move the code getting the right resolution to a function.

- Yes, I added the code for the sliders to revert back after lifting fingers. It is DualDriveSeekBar.kt, line 90:
           MotionEvent.ACTION_UP -> {
                zeroReverter.cancel()
                zeroReverter.schedule(500)
            }
Let me know if this does not work for you.
- For the python, just wondering if changing the resolution back to 1280x720 fixes the issue. I also observe the video cannot be displayed on the python sometimes, but it does not happen often.

- I think if you are happy with fixes, I do not have anything new to add at this point.

thias15:
Ok, I will merge it. So the fix for the resolution is not ideal. It also relies on the CameraConnectionFragment, which is only used in the DefaultActivity. My plan was to move all of these things to a util class.

izivkov:
> Ok, I will merge it. So the fix for the resolution is not ideal. It also relies on the CameraConnectionFragment, which is only used in the DefaultActivity. My plan was to move all of these things to a util class.

OK, no problem, I can refactor in my next PR. Thanks.

thias15:
> * OK, so the new resolution of 640x360 caused some problems. I saw your fix in RtspServer. Does it work with all phones now? 

Yes, at least all phones I tested. It will pick the supported resolution closest to the target resolution. 

> * Yes, I added the code for the sliders to revert back after lifting fingers. It is DualDriveSeekBar.kt, line 90:
>   MotionEvent.ACTION_UP -> {
>   zeroReverter.cancel()
>   zeroReverter.schedule(500)
>   }
>   Let me know if this does not work for you.

Yes works perfectly.

thias15:
> OK, no problem, I can refactor in my next PR. Thanks.

I'll do some basic refactoring and then you can take it from there. Biggest priority imho going forward is to make video stream more reliable and less latency. Also, for me when turning WiFi off on both phones Nearby connection was not always made and it does turn WiFi on automatically.

thias15 (approved):

thias15 (approved):

PR #185, opened 2021-04-30
Ivo/Misc Fixes

izivkov:
- Delay starting the RTSP client 1 second to improve staring issues.
- Move setting resolution from RtspServer to PhoneController
- Select Nearby connection when WiFi is turned off.


android/app/src/main/java/org/openbot/env/IVideoServer.java
```diff
@@ -5,10 +5,10 @@
 import android.view.TextureView;
 
 public interface IVideoServer {
-  void startServer();
```
thias15, line 8: Why is this no longer needed?

thias15: I see, this seems to be handled now in RTSP server with `private void startServer` and is not part of the interface anymore.

thias15 (commented):
Pending testing. I don't see how video is now supported in all fragments.

izivkov:
> Pending testing. I don't see how video is now supported in all fragments

Yes, the name of the branch is misleading. This was my intent, but as I described in Slack, ran into problems with dynamically creating surfaces. The description of the MR is correct. I will see if I can change the name of the brarch.

android/app/src/main/java/org/openbot/env/IVideoServer.java
izivkov: > Why is this no longer needed?

The server is started implicitly now when all the input conditions are met. Take a look at how `AndGate` is used.

izivkov: On second thought, maybe we should keep this in the interface. Looks like a method you would normally use. I think I will change it back.

thias15:
Ok, makes sense. Will test and then merge.

thias15 (approved):

PR #188, opened 2021-05-01
Model management

dhruv2295:
modal management screen to edit model details and add download it.


thias15:
- Add option to delete a model
- When adding a new model the settings dialog should show up with the default settings pre-filled
- If user selects adds a model, the file picker opens and user selects a file. We should then show the settings dialog with default settings for autopilot filled in rather than adding directly with that assumption. User can edit or confirm. This way they could also add other models in principle.

thias15:
A few things broke now. In the autopilot and object detector screen the models are not loaded correctly anymore. I guess because of the extension. Also, when you add a model from storage, the toast message still appear before clicking "Done". This is confusing. Lastly, there is an issue with Yolov4. After downloading, if I click on the model to edit it the app crashes.

thias15:
So it turns out the models did not run for me due to permissions. If I deny camera permission in another screen, it won't ask again and the models just don't work. There is another issue to clean up the permission handling across the app. Let's fix it then. For now we can assume that the user will give the camera permission.

android/app/src/main/java/org/openbot/autopilot/AutopilotFragment.java
```diff
@@ -580,7 +588,6 @@ private void setModel(Model model) {
     if (this.model != model) {
       Timber.d("Updating  model: %s", model);
       this.model = model;
-      preferencesManager.setAutoPilotModel(model.toString());
```
thias15, line 583: Would be nice to still maintain the previously selected model.

thias15 (commented):
In the autopilot screen, models can be added, but they do not show up in the model management and also not in the drop-down.

thias15 (approved):

Issue #201, opened 2021-06-03
Refactor the policy training code


marcelsan:
- [Done in #204] implement tf-records for more robust data loading.
- remove the duplicated code from Jupyter, WebServer backend, etc and use the openbot package.
- [Done in #204] upgrade to latest tensorflow version.
- make sure it works with Colab.

thias15:
I think the only thing missing now, is a Colab version of the notebook.

marcelsan:
Yes, I will do this over the weekend and create another PR. Thanks for pointing that out.

PR #204, opened 2021-06-22
Implements tf-records

marcelsan:
As suggested in #201, this PR implements a tfrecord generator scripts. Also, it modifies the training script to enable reading training and testing data from a tfrecord file.


thias15:
The subscribers are only for communication with the frontend. If we run the notebook we do not need them. 

```
2021-07-17 10:13:57.207600: E tensorflow/core/profiler/internal/gpu/cupti_tracer.cc:1661] function cupti_interface_->Subscribe( &subscriber_, (CUpti_CallbackFunc)ApiCallback, this)failed with error CUPTI_ERROR_NOT_INITIALIZED 
```

PR #205, opened 2021-07-09
Ivo/tilt control

izivkov:
- Can now drive by just tilting the controller. See updated README file in the controller directory.
- Selections on the controller for sound and image mirroring are automatically persisted on the bot app. so they are preserved across app restarts and switching controller device.
- Added a kill switch to stop bot if does not receive drive commands within 1/2 second. 
- Updated the README file.


android/app/src/main/java/org/openbot/common/ControlsFragment.java
```diff
@@ -211,6 +217,34 @@ private void handlePhoneControllerEvents() {
         );
   }
 
+  class KillSwitch {
```
thias15, line 220: Are you constantly sending some controls to make sure the bot does not get killed by mistake. For example if user drives at a constant speed, there may not be any new event for a while. 

For the communication between robot and bot app, I have implemented a heartbeat (will make a PR soon). Basically every 250ms a heartbeat message will be sent from the phone to the bot. The bot will stop if there is some issue in the communication channel.

android/app/src/main/java/org/openbot/env/NetworkServiceConnection.java
```diff
@@ -149,12 +147,7 @@ public void onServiceLost(NsdServiceInfo service) {
           ((Activity) context)
               .runOnUiThread(
                   () -> {
-                    try {
```
thias15, line 152: Why is this no longer needed?

izivkov: I moved the try/catch to the function `ControllerToBotEventBus.emitEvent()` to make the calls cleaner, instead of checking with every call.

izivkov: I moved the try/catch to the function `ControllerToBotEventBus.emitEvent()` to make the calls cleaner, instead of checking with every call.

android/app/src/main/java/org/openbot/common/ControlsFragment.java
izivkov: We send drive commands every 50 ms when the controller is active. The kill switch will trigger if it does not receive drive command in 500 ms.

izivkov: When using sliders to drive, this is usually a non-issue, because the sliders revert back to zero. Using tilt control can leave the bot running after the controller has stopped sending commands. That is why I added this kill switch.

thias15: Ok sounds good. I have tested everything and it looks good. The one thing is that it is a bit difficult to find the zero point again to stop the car. I was wondering if it makes sense to add a brake button to the screen. Maybe even do throttle control on the screen and only use accelerometer for steering. What are your thoughts on this?

thias15: As for the kill switch, let's put in into a separate file since it is a class. Or is there a way to do this from the controller? I would prefer not to have this kill switch wrapper.

izivkov: - OK, I will remove the KillSwitch. It is a bit inefficient anyway. My original idea was to have this "Transmission" class wrapper, which can change the driving characteristics, and a killSwitch() being just one method. But I see your point, I will move it to another class and not use it. The only issue is a `runaway bot`. We send a stop command on `disconnect`, but if we change the controller type on the bot app, it will not get a disconnect event. Anyway, I will check this again.

- For the zero point, I think I have a `dead zone` which will send 0,0 if the control is near zero. I will check if this is still there, and maybe widen the range. There is also an option to stop the robot if you place the controller flat on the table at any time, which should work now.

- I will update the PR soon.

izivkov: "I was wondering if it makes sense to add a brake button to the screen. Maybe even do throttle control on the screen and only use accelerometer for steering. What are your thoughts on this?"

Yes, I have seen this in other apps, like "Racing in a Car" app. If you think this is better, we can do this. Full "Zen" mode maybe a bit more difficult to drive.

thias15: I guess it's hard to know without trying. But I have also seen this in many car racing games. I guess it's easier to control only one axis via rotation.

izivkov: - I added `accelerator` and `break` pedals to the tilt screen. These are implemented in ```Accelerator.kt``` and ```Break.kt``` files. The icons are very basic, but if somebody can draw a better pedals, please go ahead and replace them.
- In both these file, there is a class called `DrivingCharacteristics`, which defined various parameters controlling responsiveness of the pedals. One can experiment with these to archive optimum driving experience.

thias15:
I like the pedal. I think it would be more intuitive if the car speed decreases when releasing the accelerator. When pressing the brake, the car should instantly go 0 and then start driving backwards. This is not like a real car, but common in car racing games on the phone.

android/controller/README.md
```diff
@@ -42,6 +50,19 @@ You can also set right/left
...
+The controller can also use its accelerometer motion sensor to drive the robot. If you select this option, the controller will enter a full-screen (Zen) mode with only 
+the video showing and no other controls. To exit this mode, either double-tap on the screen, or place the controler on a flat surface.
+
+Here is a picture of the orientation access of the controller:
+
+<p float="left">
+  <img src="../../docs/images/orientation_axis.png" width="50%" />
+</p>
+
+Start with the controller in the up-right position in landscape mode. To move forward, increase the `roll` angle (rotate forward around the `Y` access). To turn left or right,
```
thias15, line 63: The documentation needs to be updated.

izivkov (commented):
> I think it would be more intuitive if the car speed decreases when releasing the accelerator. 

- It is supposed to do this already. In the `Accelerator` class, `decelerationTime` controls how long it takes to decelerate after releasing the pedal, and `minSpeed` controls what is the final decelerated speed. Currently, `decelerationTime` is set to 5 seconds, maybe this is too long.

- I will add ability to go backwards when pressing the break. I was thinking, we should start going backwords after some delay, say 2 seconds (I will add it as param to easily change it)

I will also update the documentation, just waited to finalise these.

izivkov:
The `brake` and `accelerator` pedals could be better looking. I spent quite a lot of time on these, originally rotating one pedal 90 degrees to use of the other pedal. This created some problem where the rotated pedal disappeared after a while. If somebody wants to make a better pedal icons, please go ahead.

android/controller/src/main/java/org/openbot/controller/customComponents/Accelerator.kt
```diff
@@ -0,0 +1,130 @@
...
+        @FloatRange(
+            from = 0.0,
+            to = 1.0
+        ) val minSpeed: Float = .1f, // what percentage of full speed is our final decelerated speed (.1 means 10%)
+
+        @IntRange(
+            from = 1,
+            to = 10
+        ) val stepsToFullStop: Int = 3, // in haw many steps to get to min speed after releasing the accelerator
+    )
```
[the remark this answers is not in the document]
thias15: Looks good. Yes, the delay before going backwards makes sense.

thias15 (approved):

thias15:
I'll just update the image with a compressed jpeg.

PR #218, opened 2021-09-07
simplify load image

dvdhfnr:
`decode_image` can handle different formats and allows to direclty specify channels and dtype

### DIFF
--- a/policy/openbot/utils.py
+++ b/policy/openbot/utils.py
@@ -163,12 +163,9 @@ def list_dirs(path):
 
 
 def load_img(file_path):
-    # load the raw data from the file as a string
     img = tf.io.read_file(file_path)
-    # convert the compressed string to a 3D uint8 tensor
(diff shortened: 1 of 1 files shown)


thias15:
Did you test this? I don't see the dtype option in the docs. 
https://www.tensorflow.org/api_docs/python/tf/io/decode_jpeg

dvdhfnr:
> Did you test this? I don't see the dtype option in the docs.
> https://www.tensorflow.org/api_docs/python/tf/io/decode_jpeg

Yes. Pls note that I am using https://www.tensorflow.org/api_docs/python/tf/io/decode_image

thias15:
> > Did you test this? I don't see the dtype option in the docs.
> > https://www.tensorflow.org/api_docs/python/tf/io/decode_jpeg
> 
> Yes. Pls note that I am using https://www.tensorflow.org/api_docs/python/tf/io/decode_image

Makes sense. Thanks!

PR #227, opened 2021-09-22
on model delete, do not remove it from model management list

dhruv2295:


thias15 (commented):
This works for the object detector models where a model with download url exists. But it creates another issue. Autopilot models that were added from the server can no longer be deleted. The dialog will show up, but after deletion they still show up in the model manager. The tflite model is null though, so they cannot be used.

thias15 (approved):

PR #224, opened 2021-09-11
Update to tensorflow 2.6.0

thias15:

PR #229, opened 2021-09-27
Update dependencies

thias15:


thias15:
Revert to `tf==2.5.1`. See #224 for details.

thias15:
Tensorflow version in the app cannot be above 2.4.0 since on some phones the Yolo object detection models will not run on the GPU or cause an app crash. The Yolo object detection models need to be converted with tensorflow==2.3.0 and will use the 2.2.0 runtime. With tensorflow versions above, the select ops are required due to the `FusedBatchNormV3` and there are issue with the GPU delegate.

PR #234, opened 2021-11-12
Use GITHUB_TOKEN for actions

thias15:


thias15:
Using the [GITHUB_TOKEN](https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/) may solve this issue.

PR #247, opened 2021-12-27
Ivo/improve stability when controller is restarted

izivkov:
Androd Controller:
- Can now restart Android controller multiple times and connect to Bot without restating the Bot app.

NodeJS:
- Can reload page, close and open page without a need to restart the Bot app or the NodeJS server
- Refactored JS code to use functions instead of the less-common Classes. Streamlined code.
- Added ConnectionManager function to localise all connection-related code.


thias15 (commented):
Looks good, need to test.

thias15 (approved):

PR #249, opened 2022-01-08
Ivo/swich camera

izivkov:


izivkov:
In this PR:
- Added `button` `<span class="material-icons-outlined">
cameraswitch
</span>` to [Android] and NodeJS controllers to switch bot camera remotely.
- Minor fixes in nodejs code.
- Updated documentation.

controller/node-js/server/remote-keyboard.js
```diff
@@ -10,6 +10,17 @@
 function RemoteKeyboard (commandHandler) {
   const pressedKeys = new Set()
 
+  // This deconstruction does not work. The functions are not called. Why?
```
thias15, line 13: What is this supposed to do?

izivkov: This is not really critical, but in JS it is common to assign object properties to local variables, like this:
```
const {reset, goForward, goBackward, forwardLeft, forwardRight...} = commandHandler
```
is equivalent to:
```
const reset = commandHandler.reset
```
So, instead of calling:```commandHandler.reset()```we can just call:```reset()```

But in this case, if we call ```reset()```, the function is not called. I was wondering why, because functions are first-class citizens in JS, and should be able to use them just like any other property. Maybe somebody can answer.

izivkov:
> 

Interesting...
I looked into this a little bit deeper. It appears that we can implement our own CameraSwitchHandler by passing in into:
```((CameraVideoCapturer) videoCapturer).switchCamera(null);```
instead of passing null. Then we can select a camera by name. But I'm not sure we can do better than the default behaviour, since the camera names could change or vary across phones. What do you think?

izivkov:
> > 
> 
> Interesting... I looked into this a little bit deeper. It appears that we can implement our own CameraSwitchHandler by passing in into: `((CameraVideoCapturer) videoCapturer).switchCamera(null);` instead of passing null. Then we can select a camera by name. But I'm not sure we can do better than the default behaviour, since the camera names could change or vary across phones. What do you think?

Actually, after running through the debugger, camera names are just "0", "1" ... So we do not know if they are front, back. macro, etc. They just select them in sequence. So I don't think we can chose specific camera, since we don't know what it is.

thias15:
Should be possible to find out if it is front or rear camera. 
https://developer.android.com/reference/android/hardware/camera2/CameraCharacteristics#LENS_FACING

But I think it is also a nice feature to be able to toggle between the rear cameras. For driving it is good to use the ultra-wide camera. I guess the coolest would be if phone reports the available cameras and user can choose one. But maybe that is overkill. We could also merge now and look into this in a separate PR if you are interested to look into it.

izivkov:
Sounds good.

Ivo

controller/node-js/server/remote-keyboard.js
thias15: Check this. 
https://stackoverflow.com/questions/45910018/how-to-bind-methods-when-destructuring-an-object-in-javascript

PR #265, opened 2022-03-03
update collision thresholds

thias15:


quentin-leboutet (approved):
I confirm the proposed changes correct the bumper behaviour on the new RTR vehicle. I also did a rapid test on one of the old RTR vehicles and got -- as expected from code line 803 -- the "??" output on every bumpers, except for the "Left Back".  I approve the changes.

PR #272, opened 2022-03-31
Firmware support for RTR_520 with ESP32

thias15:


firmware/openbot_nano/openbot_nano.ino
```diff
@@ -235,6 +252,79 @@ const int PIN_PWM_R1 = 9;
 const int PIN_PWM_R2 = 10;
 const int PIN_LED_LI = 4;
 const int PIN_LED_RI = 7;
+//-------------------------RTR_520----------------------//
+#elif (OPENBOT == RTR_520)
```
quentin-leboutet, line 256: I need to redefine the min and max functions to get it to build: 
```
#define max(a, b)                   \
    (                               \
        {                           \
            __typeof__(a) _a = (a); \
            __typeof__(b) _b = (b); \
            _a > _b ? _a : _b;      \
        })

#define min(a, b)                   \
    (                               \
        {                           \
            __typeof__(a) _a = (a); \
            __typeof__(b) _b = (b); \
            _a < _b ? _a : _b;      \
        })
```

quentin-leboutet (changes requested):
The min and max functions need to be redefined so get it to build properly. Once this is done, I confirm that I can control the RTR_520 vehicle.

firmware/openbot_nano/openbot_nano.ino
```diff
@@ -251,10 +341,12 @@ int ctrl_min = (int) 255.0 * VOLTAGE_MIN / VOLTAGE_MAX;
...
 
 #if HAS_SONAR
+#if (OPENBOT != RTR_520)
 #include "PinChangeInterrupt.h"
+#endif
 //Sonar sensor
 const float US_TO_CM = 0.01715;              //cm/uS -> (343 * 100 / 1000000) / 2;
-const unsigned int MAX_SONAR_DISTANCE = -1;  //cm
+const unsigned int MAX_SONAR_DISTANCE = 300;  //cm
 const unsigned int MAX_SONAR_TIME = MAX_SONAR_DISTANCE * 2 * 10 / 343 + 1;
```
quentin-leboutet, line 350: As discussed with Matthias, an alternative solution consists in changing this line to:
```
const unsigned long MAX_SONAR_TIME = (long)(MAX_SONAR_DISTANCE * 2 * 10 / 343 + 1);
```

thias15: Updated as discussed.

firmware/openbot_nano/openbot_nano.ino
thias15: This should be fixed now with the fix below.

quentin-leboutet (approved):

PR #278, opened 2022-04-26
Multi-Terrain Vehicle (MTV)

quentin-leboutet:
Including the Multi-Terrain Vehicle (MTV) as an additional OpenBot platform.


thias15:
What are the small files like `body/mtv/cad/Assembly/._1.jpg` ?
Also, I don't think it makes sense to have duplicate copy of the phone mount in rc_truck and diy. How about we move it to the base directory?

quentin-leboutet:
Agreed.

thias15:
Great. One more things is that we should compress the images where possible. Since there are many images it will blow up the repo size. For example `docs/images/MTV/O20.pngdocs/images/MTV/O20.png` does not need to be super high-resolution and occupy 500KB. A tool like this could be used, it strikes good balance between quality and size. https://tinypng.com/ 
One option may also be to convert images to jpg.

quentin-leboutet:
Agreed. I converted the MTV pictures into .jpg

body/mtv/pcb/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+The MTV control architecture is of cascaded type. The high level commands are computed on a smartphone, running the Intel [OpenBot](https://www.openbot.org/) framework. These commands are then passed to a ESP32 low-level control board which handles PWM generation as well as communication with the different MTV sensors (e.g. encoders, sonars and so on).
+
+<p align="center">
+  <img src="../../../docs/images/MTV/Ctrl_arch.png" width="800" alt="App GUI"/>
+</p>
+
+The ESP32 low-level control board communicates with the smartphone through a dedicated serial-USB interface. The command strings have the following structure: 
+
+- | **Drive letter** | **LHS speed** | **RHS speed** |
```
thias15, line 13: This section here should probably be removed. We will control it with our OpenBot firmware. But the description of this should probably be in the main README file.

thias15 (commented):
I left a comment in the README. Also the changes to the firmware for supporting the MTV are still missing.

thias15:
The explanation about firmware in the MTV README should move to the firmware README file into a new section that explains how to use ESP32 boards. It is not specific to the MTV.

quentin-leboutet:
Makes sense.

firmware/openbot_nano/openbot_nano.ino
```diff
@@ -29,23 +30,24 @@
...
+#define RTR_520 6 // Ready-to-Run with 520-motors --> select ESP32 Dev Module as board!
+#define MTV 7     // Multi Terrain Vehicle --> select ESP32 Dev Module as board!
 
 //------------------------------------------------------//
-//SETUP - Choose your body
+// SETUP - Choose your body
 //------------------------------------------------------//
 
 // Setup the OpenBot version (DIY,PCB_V1,PCB_V2, RTR_TT, RC_CAR, LITE, RTR_520)
-#define OPENBOT DIY
```
thias15, line 45: Keep default as DIY

firmware/openbot_nano/openbot_nano.ino
```diff
@@ -74,6 +76,7 @@ boolean coast_mode = 1;
 // HAS_BUMPER                           Enable/Disable bumper (1,0)
 // HAS_SPEED_SENSORS_FRONT              Enable/Disable front speed sensors (1,0)
 // HAS_SPEED_SENSORS_BACK               Enable/Disable back speed sensors (1,0)
+// HAS_SPEED_SENSORS_MIDDLE                Enable/Disable middle speed sensors (1,0)
```
thias15, line 79: Indent should be the same as for other items.

firmware/openbot_nano/openbot_nano.ino
```diff
@@ -325,6 +328,64 @@ const int BUMPER_RF = 2930;
...
+#define HAS_LEDS_FRONT 0
+#define HAS_LEDS_BACK 0
+#define HAS_LEDS_STATUS 0
+const int PIN_PWM_R = 19;
+const int PIN_DIR_R = 18;
+const int PIN_PWM_L = 33;
+const int PIN_DIR_L = 32;
+
+// Encoder setup:
+const int PIN_SPEED_RF_A = 14;
```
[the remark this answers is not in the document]
quentin-leboutet: I agree. I corrected that.

firmware/openbot_nano/openbot_nano.ino
quentin-leboutet: Fixed

thias15 (approved):

PR #280, opened 2022-05-04
check if stream output sizes exist to avoid app crash

thias15:


thias15 (commented):
Tested by Taylor Bradley on Samsung Galaxy S10.

PR #291, opened 2022-07-29
use R1 button to discard logging

thias15:
This feature is meant to make to data collection easier. If a recording is bad, user can discard it by pressing the R1 button instead of saving the recording.


quentin-leboutet (approved):
Works as expected. I approve these changes.

PR #309, opened 2022-11-25
Integrate object detector Yolov5

thias15:
- Flip image for front camera
- Integrate Yolov5


quentin-leboutet (approved):
I confirm I can: 

- download the new yolov5 models 
- run detection on different classes, with CPU, GPU and NNAPI 

No bug was observed. I approve this PR.

Issue #310, opened 2022-11-28
Object tracking ignoring distance to object


reger24:
**Is your feature request related to a problem? Please describe.**
In object tracking mode, robot follows person with preset speed but only stops if tracked object is lost (robot w/o sonar).

**Describe the solution you'd like**
Adjust the speed by a estimated of distance. This could be implemented by measuring the box width of the tracked object.
Under the (true) assumption that tracking box width is large if close to the tracked object and small at a distance.

**Describe alternatives you've considered**
IR or sonar distance measurement could perform a emergency stop, but it's not the same as controlled slowing down.

**Additional context**
Additionally this could help to adjust speed to the movement of the tracked object

PR #311, opened 2022-11-28
Adjust robot speed by a estimated distancefactor.

reger24:
Adjust robot speed (in object tracking mode) by a distance factor estimated from the width of the tracking box
assuming that the tracking box is lage if close to tracked object and apply the factor to the speed control vars.

https://github.com/isl-org/OpenBot/issues/310


thias15:
Hi @reger24. Thanks for your contribution. Using only the width of the bounding box may be quite specific to the object of choice and the orientation of the phone. For example the ratio between bounding box width and frame width would be quite different if the phone is in landscape or portrait mode and also very different if you track a person vs. a car. I think it would probably be better to use the area (width*height) instead.

thias15:
Also, the style check did not pass. You can fix the style by navigating to the android folder and then executing `./gradlew applyStyle`

reger24:
Hi @thias15 , your point with the orientation dependency is very likely true (as I tested & concentrated on landscape).
I'll try your suggestion using the area and update my proposal/pull-request.
Thanks & Tschüß

thias15:
@reger24 sounds good. You can just push updates to your branch and the PR will be updated automatically.

reger24:
Hi @thias15 ,  https://github.com/isl-org/OpenBot/pull/311/commits/18effda2000304782289548e172a31eaac134c80 is the final change to the 1st implementation of it (if you choose so).
cu

thias15:
Hi @reger24. Sorry for the delay, very busy few months. This looks good, but since it does not work reliably in portrait mode for some classes, how about making this scaling optional? There could be a `on/off` for dynamic speed; its setting could be stored in the shared prefs. What do you think?

reger24:
> Hi @reger24. 
 There could be a `on/off` for dynamic speed; its setting could be stored in the shared prefs. What do you think?

@thias15 sounds good to me (with shared pefs switch), I'll update the pr in the next days.

thias15:
Sorry for being so picky. But ideally the dynamic speed toggle would be only in the ObjectNav GUI since it's specific to that. The settings menu is meant more for global settings like permissions, video streaming mode, robot connection, etc.

reger24:
No problem (was not completley convinced myself about the best place of the option switch), for now I'll try to squeeze it to ObjectNav's bottom sheet - and update PR.

P.S.
BUT
I believe sooner or later it would be benefitial to have a separate settings/preference screen for such option settings (which go into/are stored in openbot_settings.xml) as I foresee some more to come, to make them visible/adjustable (as preference w/o overcrowding bottomsheets) .
So I intend to prepare a **separate** PR 
- with a "Advanced (AI) Settings" or "Detail AI Setting" which make parts of openbot_settings.xml visible
- add a Link to the "Settings" which opens the Advanced Settings screen, with categories for "General (all AI screens)" (eg. CAMERA_SWITCH) and individual (eg. "Object Tracking") for related settings.
  - later (if there are more options)... a button on bottomsheet could also open the Advanced Settings screen

thias15 (approved):

PR #314, opened 2022-12-06
Quentin leboutet/point goal nav training

quentin-leboutet:
Integration of the point goal navigation training pipeline


thias15:
Please fix style and test if Autopilot still works the same.

thias15 (approved):

PR #317, opened 2022-12-10
Add efficientDet models

thias15:
- add efficientDet models
- refactor model implementation and naming
- check for internet connection in model management fragment


quentin-leboutet:
Model management and autopilot appear to work flawlessly. Object tracking also works well for every available model except for the efficientDet with NNAPI. In this case, I observed two different behaviors: 

1. either the indicated fps is similar to that of the CPU mode but nothing is being detected
2. or that the fps drops to 0.5-1.5 fps with some successful detections

thias15:
@quentin-leboutet, I added a remark about landscape mode in combination with the EfficientDet models. Also added some benchmarking results. NNAPI performance highly depends on the phone. Maybe you can also test on an additional phone. If there is nothing else, please approve.

PR #318, opened 2022-12-17
Prep deletion of redundant code getScreenOrientation

reger24:
The code in CameraActivity is redundant with same methode in ImageUtils and only 1 time used.
Could be deleted without risk.
In this commit I just marked it deprecated (for later deletion)


thias15:
@reger24, yes you are right; good catch! You can just delete the code. No need to keep deprecated/duplicated code around. Thanks!

reger24:
> can just delete the code

Sure, will do (in a minute)

thias15 (approved):

PR #329, opened 2023-01-17
Adding iOS App Support

sparsh3dwe:
- Adding iOS app support for OpenBot.
- Written in swift.
- Dual theme (Light/Dark).
- Supports Mobile Controller as well as Remote Controller with MFI enabled.


thias15:
Hi @sparsh3dwe. There seems to be a lot of files such as `xcuserdata` that don't need to be there. Could you please check which files are build/temporary files not needed for compilation and remove them. We should only have the necessary source code.

.gitignore
```diff
@@ -6,6 +6,10 @@ composer.lock
 .DS_Store
 Thumbs.db
 wandb/
+/ios/OpenBot/Pods/
```
thias15, line 9: There should be another .gitignore file at the level `ios/OpenBot` that has `Pods`, etc. in it. Does not make sense to put this at the root level.

firmware/openbot_nano/openbot_nano.ino
```diff
@@ -116,7 +118,7 @@ const int PIN_PWM_R1 = 9;
 const int PIN_PWM_R2 = 10;
 const int PIN_SPEED_LF = 2;
 const int PIN_SPEED_RF = 3;
-const int PIN_VIN = A7;
+//const int PIN_VIN = A7;
```
thias15, line 121: Why is this disabled? The changes here should not be part of the PR.

firmware/openbot_ble/openbot_ble.ino
```diff
@@ -0,0 +1,1856 @@
+// Required App Version: 0.5
```
thias15, line 1: How many changes are required to the firmware? I think it would make more sense to integrate it with the existing firmware and use a flag such as #HAS_BLUETOOTH to turn the feature on/off in the firmware.

thias15:
`ios/OpenBot/OpenBot.xcodeproj/project.xcworkspace/xcuserdata` is in .gitignore but files like `nitishyadav.xcuserdatad/UserInterfaceState.xcuserstate` are still included in this branch.

firmware/openbot_ble/openbot_ble.ino
sparsh3dwe: Hi, I've merged the both BLE with the BLUETOOTH flag.

thias15:
@quentin-leboutet how is the testing going?

PR #330, opened 2023-01-20
Adding BLE support on Android Device + DPAD controller Connection

sparsh3dwe:


thias15:
`> Task :app:testDebugUnitTest FAILED`

sparsh3dwe:
> `> Task :app:testDebugUnitTest FAILED`

added null checks to fix the failures

thias15:
Added a few more fixes. Also noticed that the jpeg images for the README are not pushed to this PR. Please add them after [compressing them](https://tinyjpg.com/).

thias15 (commented):
README needs images and potentially some adjustments. Still need to test functionality - requires firmware from #329

android/app/src/main/java/org/openbot/autopilot/AutopilotFragment.java
```diff
@@ -275,6 +275,8 @@ private void recreateNetwork(Model model, Network.Device device, int numThreads)
 
   @Override
   public synchronized void onResume() {
+    croppedBitmap = null;
+    tracker = null;
```
thias15, line 279: Why are these set to null here?

android/app/src/main/java/org/openbot/vehicle/Vehicle.java
```diff
@@ -392,9 +392,11 @@ public boolean isUsbConnected() {
   private void sendStringToDevice(String message) {
     if (getConnectionType().equals("USB") && usbConnection != null) {
       usbConnection.send(message);
+      System.out.println("sanjeev usb = " + message);
```
thias15: Please don't push debug messages.

android/app/src/main/java/org/openbot/autopilot/AutopilotFragment.java
sparsh3dwe: So, there was a bug into the application, which is being fixed here by reinitializing the variables.

Bug was if you turn on the auto mode into the autopilot, then move to settings by clicking on the usb/ble icon and come back to the fragment. 

it crashes.

android/app/src/main/java/org/openbot/vehicle/Vehicle.java
sparsh3dwe: okay, will keep in mind

android/app/src/main/java/org/openbot/autopilot/AutopilotFragment.java
thias15: Great 👍. One 🐛 less!

thias15 (commented):
Looks good, tested. Just need to adjust README with updated text and screenshots.

PR #332, opened 2023-02-05
Get DEFAULT camera lens facing from shared preferences,

reger24:
**Is your feature request related to a problem? Please describe.**
Some (me) use robot app front camera for object tracking etc.. 
This has to be manually toggled each time the app is started.

**Describe the solution you'd like**
Would like to define the DEFAULT camera lens facing direction to use in the preferences.

**Description of changes**
Pull request implements a shared preference setting for the default camera facing.
It leaves the bottom sheet toggle camera switch as is, so camera can be switched manually as usual without altering the preference setting.
On next start of the robot app the default from shared preferences is used.

### DIFF
--- a/android/app/src/main/java/org/openbot/common/CameraFragment.java
+++ b/android/app/src/main/java/org/openbot/common/CameraFragment.java
@@ -27,6 +27,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import org.openbot.R;
 import org.openbot.env.ImageUtils;
+import org.openbot.env.SharedPreferencesManager;
 import org.openbot.utils.Constants;
 import org.openbot.utils.Enums;
 import org.openbot.utils.PermissionUtils;
@@ -38,7 +39,7 @@ public abstract class CameraFragment extends ControlsFragment {
   private ExecutorService cameraExecutor;
   private PreviewView previewView;
   private Preview preview;
-  protected static int lensFacing = CameraSelector.LENS_FACING_BACK;
+  protected int lensFacing = CameraSelector.LENS_FACING_BACK; // altered by shared preferences
   private ProcessCameraProvider cameraProvider;
   private Size analyserResolution = Enums.Preview.HD.getValue();
   private YuvToRgbConverter converter;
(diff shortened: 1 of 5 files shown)


thias15:
It does not work for me. Tested on Pixel 4 XL. If I toggle the camera and then go to a different fragment or restart the app, it defaults back to the back camera.

reger24:
Hm, makes me wounder that it is not working for you (maybe you overlooked the Settings screen),
but anyway....
looking deeper into the usage of the prefs leaves me a bit confused about what is intended (see below) to replace PR.

Currently the app creates two preference/setting file
- **org.openbot_preferences.xml** created by Settings screen
- **openbot_settings.xml** created by class SharedPreferenceManager and used & altered by DefaultActivity's - BottomSheet (while other BottomSheets do NOT alter, only read the preferences)

and at least on value is **redundant** (usb **baud_rate** with different keys ("BAUD_RATE" vs. "baud_rate") ond value.types (string vs. int))

This seems **inconsistent** as well as the **behavior** of bottomsheet effecting stored preference (**DefaultActivity**.Bottomsheet **reads&writes** prefs, **other** Bottomsheets **only read** prefs).

For consistency I'd like to workout a new proposal/PR and the **here proposed setting could** likely **use the existing CAMERA_SWITCH** pref (currently only used by DefaultActivity).

But I'm **not sure about your pref** & the best solution, 
imho options are:
- [ ] **a)** use just one preference xml file (and move the other prefs over)
- [ ] **b)** leave both but define what should go where, e.g.
     - org.openbot_preferences.xml (Settings screen) just for application related general settings (color, styles, permission etc)
     - openbot_settings.xml (SharedPreferenceManager) for all bot feature detail setting/option (currently only used programatically by bottomsheet)
     - create **additional "Bot Setting" screen** to view and alter settings (as bottomsheets may get crowded over time)
     - all **bottomsheets read&write** prefs

**?**

thias15:
Yes good point. At the moment there are two different preferences, one for the settings menu and one for default values across the other app screens. I agree, it would be best to use the existing camera_switch pref which should actually work across fragments (works for per per session). I guess I might have not tested your PR correctly, I did not realize that you added something to the settings menu. Can check it again. But I guess in the current case the selection in the settings screen would then override whatever changes the user makes while using the app.

PR #333, opened 2023-02-07
Upgrading Android API Level from 29 to 32.

sparsh3dwe:
- Upgrading android api level from 29 to 32.
- Changing default logs and files output location to Documents.
- Upgrading Gradle version from 7.0.4 to 7.4.0


sparsh3dwe:
> Looks good. Need to test. Can recordings be accessed directly from file storarge?

Yes, they are directly accessible from documents

thias15:
Tested, works as expected. The only issue is that the app is not able to find the node.js server for uploading data. But this seems to be unrelated to this PR.

thias15 (approved):

PR #359, opened 2023-05-21
Node.js Web controller client UI improvements

Coder-Lars:
Screenshot_1
Screenshot_2


thias15:
New UI looks nice. Lots of files needed for the new font. Any other option? Minor comments: `Controls` should be spelled with only one `l`

thias15:
Thanks, for the quick response. A few more minor comments.
1) I think it would look nicer with equal button size (for single letter keys). What do you think?
2) For left/right/up/down you could consider using the corresponding arrow symbols like on the keyboard.
3) Please update the README.md by replacing the current screenshot of the UI with one of yours.

thias15:
Great work @Coder-Lars!

thias15 (approved):

PR #373, opened 2023-09-04
Bug Fix: Tflite files are not uploading in Model Management on Android Devices

sparsh3dwe:


quentin-leboutet (changes requested):
I cannot delete a newly loaded model within the model management fragment. 
The functionality of adding new tflite models also seems to be broken.

quentin-leboutet (approved):

Issue #385, opened 2023-12-06
ObjectNavSegment control result should be mirrored when using front facing camera


yijunwu:
**Description**
Currently, in ObjectNavSegment, when using front facing camera of the phone, the vehicle's target control is incorrect. When the object is at the vehicle's left front, it turns to right, and vice versa. Currently, in order for the bot to do object nav correctly, one has to put the phone upside down, but consequently the controling UI is upside down too, which is inconvienient to use.

**To Reproduce**
1. With current release 0.7.1, let the phone screen facing to vehicle's front, and switch to front facing camera
2. Go to Object Tracking segment
3. Try object tracking
4. See error - control should be mirrored

**Expected behavior**
When the phone screen facing to vehicle's front, and using front facing camera, control target should be correct so that the bot can follow detected object.

**Smartphone:**
 - Device: some main stream phones, Samsung S20FE & Samsung A40S tested
 - OS: Android 10 & Android 13 tested
 - Version 0.7.1, master

**Suggested fix**
Fix should be easy, I'll create a pull request soon

yijunwu:
I've created a pull request ([386](https://github.com/isl-org/OpenBot/pull/386)) to fix this issue. I've tested the fix and it works.

PR #386, opened 2023-12-06
Mirror the target control of ObjectNavFragment if front facing camera is used

yijunwu:
When front facing camera is used, the target control of ObjectNavFragment should be mirrored otherwise the bot can't follow the target object. See the issue [385](https://github.com/isl-org/OpenBot/issues/385) for detail


thias15:
The convention for the front camera varies across phones. How about adding an icon to mirror controls next to the camera flip icon.

yijunwu:
OK, I haven't tried phones other than Samsung's. Maybe a better solution would be to detect the convention for the phone automatically, and set default value accordingly, in addition to adding icon to mirror control? Not sure if it's detectable. Before figuring out that, adding icon to mirror control can be a basic solution. I'll add some more commits soon

yijunwu:
I've added more commits. It should be ok now.
image
image

yijunwu:
I adjusted the layout and appearance of icons, now it looks like this: 
6830d259c0e1cca88ef2e2368520496
fa1deaea6ead352cafed02663e759dd

desy0305 (approved):
ok

thias15 (approved):

hardikgarg02:
I have fixed the UI for the fragment. Could you review and replace below file named `fragment_object_nav.xml` in the path `robot/src/main/res/layout-land` directory.
[fragment_object_nav.xml.zip](https://github.com/isl-org/OpenBot/files/13784523/fragment_object_nav.xml.zip)

And also replace below file named `fragment_object_nav.xml` in the path `robot/src/main/res/layout` directory.
[fragment_object_nav.xml.zip](https://github.com/isl-org/OpenBot/files/13784567/fragment_object_nav.xml.zip)

PR #433, opened 2024-06-17
Update firmware readme.md

hexs00si:
Added step 7 to the readme file in the firmware directory , earlier some users might face some error as the default upload speed is set to 921600 , but certain esp32 chips would throw error while uploading the sketch , Added the basic instruction to simple change the upload speed


thias15:
Please remove the white space and * --> _ changes. Also shouldn't it be step 4 rather than step 7?

hexs00si:
I will and yea it should be step 4 , i will most definitely do that .

hexs00si:
@thias15 , please review the updated change , i changed the instruction from instruction 7 to 4

thias15 (approved):

PR #436, opened 2024-06-18
On stopping auto mode in autopilot, the drive mode gets enable for phone controller i.e dual drive

hardikgarg02:


thias15:
Please resolve conflicts.

android/robot/src/main/java/org/openbot/autopilot/AutopilotFragment.java
```diff
@@ -463,12 +465,10 @@ protected Model getModel() {
 
     private void connectWebController() {
         phoneController.connectWebServer();
-        Enums.DriveMode oldDriveMode = currentDriveMode;
```
thias15, line 462: I really don't like this code. But removing it means it won't remember the drive mode when we go back to game controller mode.

thias15:
Closing this — apologies for the long delay before getting to it.

The guard that disables the drive-mode toggle while a phone/web controller is connected is the right idea, but the way this PR achieves it introduces a regression that outweighs the fix:

1. **Drive-mode preference is silently lost.** Removing the `oldDriveMode = currentDriveMode` capture and the `setDriveMode(oldDriveMode.getValue())` restore means once a user connects a phone/web controller, their original drive-mode preference is overwritten in `SharedPreferences` and never restored on disconnect. They're stuck on `DUAL`/`GAME` until they manually change it.
2. **`CameraActivity.connectPhoneController` now declares `DriveMode oldDriveMode = currentDriveMode;` but never reads it** — dead variable, compiler warning.
3. **Guard logic is duplicated** between `AutopilotFragment.setNetworkEnabled` and `ObjectNavFragment.setNetworkEnabled` (same 7-line block copy-pasted) — would drift over time.
4. **Conflicts with #498** (which replaces `currentDriveMode` reads with `vehicle.getDriveMode()` in the same methods) and with current master.

If you'd like to revisit, a clean version would: keep the new guard in `setNetworkEnabled` (extracted to `ControlsFragment` so it isn't duplicated), and properly restore the user's prior drive mode in `disconnectPhoneController` instead of removing the save+restore entirely. Happy to review a fresh PR.

PR #464, opened 2024-11-25
Flutter controller v2

isha382:
This is having new changes in ui for flutter controller which is now having a setting icon in which there is option to change the controller mode and some other ui changes.


docs/images/app_controller_settings_1.jpg
thias15, line 1: Do we need those?

thias15:
Sorry for the very long delay on this review. Submitted my pending comment above; here's the broader feedback:

1. **Dead images.** `docs/images/app_controller_settings_1.jpg` and `app_controller_settings_2.jpg` aren't referenced anywhere in the README — only `flutter_controller_setting.jpg` is. Were they intended to illustrate something that got dropped?
2. **Deprecated `import android.app.Fragment`** in `SharedPreferencesManager.java`. The codebase otherwise uses `androidx.fragment.app.Fragment`, and the import looks unused (only the string return type from `getFragment()` is involved). Could you remove it?
3. **Enum naming convention.** `Enums.fragmentType` should be `FragmentType` to match the existing `SensorType`, `LogMode`, `ControlMode`.
4. **Persistence layer that does nothing.** `SharedPreferencesManager.setFragment()` is called in each fragment's `onViewCreated` and immediately read back via `getFragment()` to emit the event. It's never used across launches — should this just be a local variable (or have the fragment type passed directly to `createFragment`)?
5. **Removed "double-tap to go back" behavior in `controlSelector.dart`.** The old code had `onDoubleTap` callbacks on `TiltingPhoneMode`/`OnScreenMode` to flip back to mode selection. The new code wraps the same widgets in `GestureDetector` *without* an `onDoubleTap`, so the gesture detectors are dead code. Was the intent to drop the gesture in favor of the new menu drawer? If so, the empty `GestureDetector` wrappers can go too.
6. **Protocol inconsistency.** Existing messages use `{"status": {"<NAME>": "<value>"}}`. The new fragment-type message uses a top-level `{"FRAGMENT_TYPE": "<value>"}`, so the Flutter side now has to branch on two shapes. Could `FRAGMENT_TYPE` be nested under `status` like the rest, to keep the protocol uniform?
7. **Conflicts with #507.** That PR also touches Flutter Android Kotlin (it goes to 2.1.0 vs 1.7.0 here) and the iOS `Info.plist` scene-manifest. Whichever lands first will force the other to rebase. Worth coordinating with the #507 author.

thias15:
Thanks for the updates — good progress. A few things still to clarify:

1. **Dead images** — looks resolved, the README now references `flutter_controller_setting.jpg` correctly.

2. **Deprecated `import android.app.Fragment` in `SharedPreferencesManager.java`** — still there? The diff only removes a blank line. Could you confirm this was removed?

3. **Enum naming** — looks resolved, `FragmentType` is now correctly capitalized.

4. **`setFragment()`/`getFragment()` in `SharedPreferencesManager`** — is this still in there? It was being called and immediately read back without serving any real persistence purpose. If so, could you remove it and pass the fragment type directly to `createFragment()`?

5. **Dead `GestureDetector` wrappers** — looks resolved.

6. **Protocol inconsistency** — `FRAGMENT_TYPE` still uses a flat `{"FRAGMENT_TYPE": "..."}` shape while all other messages use `{"status": {"NAME": "value"}}`. The Flutter side now has to handle two different message shapes. Was this intentional, or could `FRAGMENT_TYPE` be nested under `status` to keep things consistent?

7. **Conflicts with #507** — we plan to merge #507 first, so please rebase on master once that lands.

Finally — has this been tested end-to-end on both Android and iOS, including the new settings drawer and all three controller modes?

harishthakur52:
> Thanks for the updates — good progress. A few things still to clarify:
> 
> 1. **Dead images** — looks resolved, the README now references `flutter_controller_setting.jpg` correctly.
> 2. **Deprecated `import android.app.Fragment` in `SharedPreferencesManager.java`** — still there? The diff only removes a blank line. Could you confirm this was removed?
> 3. **Enum naming** — looks resolved, `FragmentType` is now correctly capitalized.
> 4. **`setFragment()`/`getFragment()` in `SharedPreferencesManager`** — is this still in there? It was being called and immediately read back without serving any real persistence purpose. If so, could you remove it and pass the fragment type directly to `createFragment()`?
> 5. **Dead `GestureDetector` wrappers** — looks resolved.
> 6. **Protocol inconsistency** — `FRAGMENT_TYPE` still uses a flat `{"FRAGMENT_TYPE": "..."}` shape while all other messages use `{"status": {"NAME": "value"}}`. The Flutter side now has to handle two different message shapes. Was this intentional, or could `FRAGMENT_TYPE` be nested under `status` to keep things consistent?
> 7. **Conflicts with [Dependency updates to latest versions. #507](https://github.com/ob-f/OpenBot/pull/507)** — we plan to merge [Dependency updates to latest versions. #507](https://github.com/ob-f/OpenBot/pull/507) first, so please rebase on master once that lands.
> 
> Finally — has this been tested end-to-end on both Android and iOS, including the new settings drawer and all three controller modes?

For points 2 and 4, those have already been removed and updated with the correct implementation.

For point 6, it's sending {"FRAGMENT_TYPE": "...."} for the home page. When the app reaches the home page, the network and Data Collection toggle buttons are disabled in controller.

We have rebase the PR and also created controller builds in both IOS and Android. They are working fine as expected

Thanks

thias15:
Thanks for the update, and for clarifying the other points. On (6), one more nudge: I don't think this needs a new envelope shape at all.

Looking at how `"status"` is already used elsewhere, it's not narrowly "current robot state", it's the generic envelope for any named bot→controller message (`VIDEO_SERVER_URL`, `WEB_RTC_EVENT` are already wrapped there despite not being "statuses" either). So `FRAGMENT_TYPE` fits the same pattern:

- In `ConnectionUtils.java`, swap `createFragment()`/`closeFragment()` for the existing `createStatus("FRAGMENT_TYPE", value)` / `createStatus("FRAGMENT_TYPE", "CLOSE")`.
- In `controller.dart`, drop the separate `if (msgInObject["FRAGMENT_TYPE"] != null)` check and instead add `FRAGMENT_TYPE` as one more field inside `processMessageFromBot`, next to `CONNECTION_ACTIVE`, `VIDEO_PROTOCOL`, etc.
- Same idea on the iOS side wherever the equivalent dispatcher lives.

What do you think? Unless you disagree, can you make that change and do a quick test. Then we're ready to merge.

harishthakur52:
> Thanks for the update, and for clarifying the other points. On (6), one more nudge: I don't think this needs a new envelope shape at all.
> 
> Looking at how `"status"` is already used elsewhere, it's not narrowly "current robot state", it's the generic envelope for any named bot→controller message (`VIDEO_SERVER_URL`, `WEB_RTC_EVENT` are already wrapped there despite not being "statuses" either). So `FRAGMENT_TYPE` fits the same pattern:
> 
> * In `ConnectionUtils.java`, swap `createFragment()`/`closeFragment()` for the existing `createStatus("FRAGMENT_TYPE", value)` / `createStatus("FRAGMENT_TYPE", "CLOSE")`.
> * In `controller.dart`, drop the separate `if (msgInObject["FRAGMENT_TYPE"] != null)` check and instead add `FRAGMENT_TYPE` as one more field inside `processMessageFromBot`, next to `CONNECTION_ACTIVE`, `VIDEO_PROTOCOL`, etc.
> * Same idea on the iOS side wherever the equivalent dispatcher lives.
> 
> What do you think? Unless you disagree, can you make that change and do a quick test. Then we're ready to merge.

Thanks for the nudge — agreed, status is already the general envelope (same as VIDEO_SERVER_URL, WEB_RTC_EVENT), so no need for a new shape. Made the change:

  - Android: createFragment()/closeFragment() removed, replaced with createStatus("FRAGMENT_TYPE", value) / createStatus("FRAGMENT_TYPE", "CLOSE").
  
  - Flutter: Dropped the separate FRAGMENT_TYPE check — now just another field inside processMessageFromBot, next to CONNECTION_ACTIVE/VIDEO_PROTOCOL.
  
  - iOS: Same pattern — FragmentTypeEvent now wraps its payload under status.

  Tested end-to-end on Android and iOS across all three controller modes — no issues. Ready to merge from my side.

  Please let me know if you'd like me to make any further changes. Thanks!

thias15 (approved):
`FRAGMENT_TYPE` now uses the shared `status` envelope on all platforms, matching the protocol. CI green. Approving, thanks for the work on this.

PR #468, opened 2024-12-14
arduino-esp32 v3.0.0 change paramter types ble

nanotuxi:
See [https://github.com/espressif/arduino-esp32/blob/master/docs/en/migration_guides/2.x_to_3.0.rst#ble](url)


nanotuxi:
This should happen im my own fork and can be deleted. Sorry for that.

PR #471, opened 2025-01-17
Feature/multi language support

kellyy0:


thias15:
Sorry it's taken so long for this to get a real review — and thanks for tackling i18n, it's overdue.

Unfortunately the implementation has a few correctness bugs that I think need addressing before this can land. The general principle is: **once you localize, never compare or store UI strings as control flow** — only compare resource IDs.

1. **`MainFragment.onItemClick` uses raw English string literals** in its switch (`case "Projects":`, `case "Free Roam":`, …). After this PR, `subCategory.getTitle()` returns the localized string, so navigation will only work in English — switching to French/German/Chinese silently breaks navigation. The switch should compare resource IDs (e.g. `subCategory.getTitleResId() == R.string.projects`) rather than strings.
2. **`SubCategory` constructor stringifies the resource ID:** `this.title = String.valueOf(title)` produces `"2131099672"` instead of a localized string. Anywhere `getTitle()` is read as a UI label will show gibberish digits. `SubCategory` should store the int resource ID and expose `getTitle(Context)` like `Category` now does.
3. **`OpenBotApplication.refreshAppLanguage`:** the static `context` field holds the Application context, so `context instanceof Activity` is always false and `activity.recreate()` never runs. Could you wire this through the actual Activity (e.g. pass it as a parameter from the settings fragment) so the language change takes effect immediately?
4. **`android/build.gradle` downgrades `com.google.gms:google-services` from 4.3.15 → 4.3.10** — unrelated to i18n and should be reverted.
5. **`MANETTE` enum value (French for "controller") leaked into the `ControlMode` enum.** Enum *names* should stay English; only the *displayed* string varies by locale. If `MANETTE` is meant to be a different control mode than `GAMEPAD`, it needs an English name; if it's just the localized label for `GAMEPAD`, the enum value shouldn't exist at all and only the `string.xml` translations need to differ.

Could you also confirm the PR has been tested end-to-end on a real device with at least one non-English locale (start the app, open every feature category, switch language at runtime, navigate everywhere) before re-requesting review? The current state would crash / show garbled text in several places.

thias15:
Thanks for taking this on! I've merged/superseded this work in #526, which extends the language coverage (adding Spanish, Korean, and Hindi alongside German, French, and Chinese) and adds an in-app language picker using AndroidX's per-app language API, along with a Settings screen redesign. Closing this one in favor of that, appreciate the contribution!

PR #481, opened 2025-02-03
updated dependency and deleted package-lock.json

isha382:
updated dependency and deleted package-lock.json


thias15:
Sorry for taking so long to look at this — and thanks for the contribution!

A larger PR (#507) is now open and approved that updates the same `controller/web-server/package.json` and `server/package.json` files much more comprehensively (and also touches the rest of the project's dependency stack). Once #507 lands, the `package.json` changes here will be subsumed.

Two questions before deciding:

1. Are you okay if we **close this in favor of #507**, or is there a specific reason you'd prefer this PR's narrower change to land first?
2. The `ws ^7.5.0 → ^8.18.0` jump on the server side is a major version bump — did you verify the websocket connection between the server and the OpenBot controller still works after the upgrade? (ws v8 changed some defaults around perMessageDeflate and event handling.)

If you'd like to keep this open as a fallback in case #507 stalls, that's fine — but in that case we'd want answers to #2.

harishthakur52:
@thias15 we can close this PR as #507 contains the same updated changes.

PR #485, opened 2025-02-13
Upload build track instructions readMe

Noobrekter69:


thias15:
Please clean up the links. Usually anything after /dp/<item_id>/ can be removed. 

For example, below everything after dp/B0CXCX684S is not needed. 

So instead of
https://www.amazon.co.uk/Corrugated-Plastic-Outdoor-Waterproof-Coroplast/dp/B0CXCX684S/ref=sr_1_33?dib=eyJ2IjoiMSJ9.qxDcKr-bIssyror76QR25zwVhHfoXiv_69JdU1qLB_h7KHJ6AbzDy35rQDBmeEErshpSuW2MgaD-96we9Xou-9-tiCoDqbz3mOsvLk3vu8SDmWN7XUN9N3dZQo5mBD1rbTQcvlNwGIEWRJS8xr2FBCtXzE40PPJ-gF3RFl9CHtV2vB9sVT0hsvRvQSnJmgqL1Zr9hNc0ORVr7E_5TyqZCg9BzYVOjEBO17quffB368B4lZ3czD4DBDH0_x4rFEvKCBQwkDyLPhRgNKRS_TT8qYVvQ96pZioXSLzKDcWL4OOUstFqym6vOfiv0xftV0DKeLxPW5IyNlP75RA56PH7y_P9tI3JIIeWM5DlvzwOp-hqX5HYSHGFOrUDuFYvHhWTrl2EyCoicjNNmThiwVPAhDTMqEttXmb_OaPM3VEoedMK14cgQTnHcBxtCASl0Kgj.cRMT8sWsb09UfjY0p6A54jI05M71OFk2c4vSODMJIeE&dib_tag=se&keywords=corrugated%2Bplastic%2Bsheets%2B53cm%2Bx%2B76cm&nsdOptOutParam=true&qid=1736752683&sr=8-33&th=1

You can do this.
https://www.amazon.co.uk/dp/B0CXCX684S

Also you need to add the images to the PR and reference them with relative links.

Noobrekter69:
@thias15 I have cleaned up the hyperlink already and added the images to PR and reference it with relative links.

thias15:
I noticed several links are the same for US/EU/ASIA. Is this intentional?

Noobrekter69:
Sometimes I cannot access US/UK's amazon website it will redirect me back to SG's amazon

thias15:
Sorry for the long delay on this — and thanks for putting these instructions together, the step-by-step photos are great.

A few things would need to be sorted before merging:

1. **Folder location:** `Build-Track-Instruction-main/` looks like the auto-generated name from a GitHub zip download. Could you move the README into the existing `body/` tree — for example `body/track/README.md` — so it lives alongside the other build instructions? The current folder name doesn't match repo convention.
2. **Image paths will be broken.** From inside the new folder, `docs/images/track_setup_Step1.0.jpg` resolves to `<your-folder>/docs/images/...` which doesn't exist. Once the README is moved, the paths will need to use the right relative prefix (e.g. `../../docs/images/...` if it ends up under `body/track/`). Could you double-check by previewing the rendered file on GitHub before requesting re-review?
3. **Empty link placeholders:** the line `Glue, tape, or Velcro for temporary attachment ([US](), [EU](), [ASIA]())` has empty URLs — please either fill them in or remove the placeholders.
4. **Region duplication:** "Metal Straightedge" has US and EU both pointing to `B07R9MW35Q`, which is actually the tape link. I understand from your earlier reply that some Amazon regions auto-redirect for you locally, but the links shipped to users should genuinely point to the right region's listing — could you ask someone with US/UK Amazon access to verify, or at least pick one canonical link and not duplicate it across all three regions?
5. **Discoverability:** could you also add a link to this new README from the top-level `README.md` (or wherever the project lists build instructions) so users can find it?

PR #498, opened 2025-06-05
Unify the driveMode storage and fetching - from ControlsFragment.vehicle.driveMode, instead of ControlsFragment.currentDriveMode.

yijunwu:
Unify the driveMode storage and fetching - from ControlsFragment.vehicle.driveMode, instead of ControlsFragment.currentDriveMode.

There are two places containing drive mode: ControlsFragment.currentDriveMode, and ControlsFragment.vehicle.driveMode. They should be unified. ControlsFragment.vehicle.driveMode is used more widely(as shown below) and is supposed to be the final one.

image

### DIFF
--- a/android/robot/src/main/java/org/openbot/autopilot/AutopilotFragment.java
+++ b/android/robot/src/main/java/org/openbot/autopilot/AutopilotFragment.java
@@ -459,7 +459,7 @@ public class AutopilotFragment extends CameraFragment {
 
     private void connectWebController() {
         phoneController.connectWebServer();
-        Enums.DriveMode oldDriveMode = currentDriveMode;
+        Enums.DriveMode oldDriveMode = vehicle.getDriveMode();
         // Currently only dual drive mode supported
         setDriveMode(Enums.DriveMode.GAME);
         binding.controllerContainer.driveMode.setAlpha(0.5f);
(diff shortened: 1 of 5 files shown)


thias15:
Sorry for taking so long to look at this — thanks for your patience.

1. **Is `vehicle.getDriveMode()` guaranteed to return a non-null value at every read site?** The old `currentDriveMode` was initialized to `Enums.DriveMode.GAME` at field declaration, so it was never null. After the refactor, the first `handlePhoneControllerEvents()` call (status reporting in `ControlsFragment`) and the `connect*Controller()` methods rely on the vehicle already having a drive mode set. If `Vehicle` initializes its `driveMode` to a sane default, this is fine — could you confirm?
2. The PR has been open since June 2025 — has it been re-tested against current master? Touching `ControlsFragment` and four downstream fragments has broad surface area; given the time gap, it'd be worth a quick verification that all four control modes (free roam, object nav, logging, autopilot) still behave correctly with both phone and web controllers.
3. **Minor (optional):** while you're touching `FreeRoamFragment.java`, the file mixes `Enums.DriveMode` and `DriveMode` (static import) — Copilot pointed this out. Not a regression from this PR, but a 1-line cleanup if you want it.

PR #507, opened 2026-03-12
Dependency updates to latest versions.

hardikgarg02:


thias15:
Looks good.

thias15 (approved):

thias15:
Was this PR fully tested after the latest changes and is ready to be merged?

harishthakur52:
Yes, tested for all required cases and dependency’s

thias15:
1. **`Runner.xcodeproj/project.pbxproj` commits a hard-coded `DEVELOPMENT_TEAM`** — won't this break iOS code signing for everyone not on that Apple team? For an open-source project, would it make sense to leave it unset (or move to a local xcconfig) so signing is per-developer?
2. **`controller/web-server/package-lock.json` and `server/package-lock.json` still reference the pre-update versions** (e.g. `debug ~2.6.9`, `firebase ^10.1.0`). Should these be regenerated and committed, or is the intent to drop the lockfiles entirely? Right now `npm ci` would resolve a different graph than `npm install`.
3. **`controller/flutter/android/app/build.gradle` replaces `flutter.versionCode` / `flutter.versionName` with hard-coded `1` / `"1.0"`.** The robot and main controller apps derive their version from git tags via the `io.github.reactivecircus.app-versioning` plugin in their `build.gradle` — could we apply the same plugin here so the Flutter controller's version stays in sync with releases? Otherwise this app would be the only one without a versioning hook.
4. **`minSdkVersion` 21 → 24 drops support for Android 5/6 users on the Flutter controller**, while the main `robot` and `controller` apps still target `minSdk 21`. Was this intentional, and should the other apps move in lockstep?

Note: one of Copilot's review comments is stale — the workflow already has the Stop-server cleanup step it was suggesting.

harishthakur52:
> 1. **`Runner.xcodeproj/project.pbxproj` commits a hard-coded `DEVELOPMENT_TEAM`** — won't this break iOS code signing for everyone not on that Apple team? For an open-source project, would it make sense to leave it unset (or move to a local xcconfig) so signing is per-developer?
> 2. **`controller/web-server/package-lock.json` and `server/package-lock.json` still reference the pre-update versions** (e.g. `debug ~2.6.9`, `firebase ^10.1.0`). Should these be regenerated and committed, or is the intent to drop the lockfiles entirely? Right now `npm ci` would resolve a different graph than `npm install`.
> 3. **`controller/flutter/android/app/build.gradle` replaces `flutter.versionCode` / `flutter.versionName` with hard-coded `1` / `"1.0"`.** The robot and main controller apps derive their version from git tags via the `io.github.reactivecircus.app-versioning` plugin in their `build.gradle` — could we apply the same plugin here so the Flutter controller's version stays in sync with releases? Otherwise this app would be the only one without a versioning hook.
> 4. **`minSdkVersion` 21 → 24 drops support for Android 5/6 users on the Flutter controller**, while the main `robot` and `controller` apps still target `minSdk 21`. Was this intentional, and should the other apps move in lockstep?
> 
> Note: one of Copilot's review comments is stale — the workflow already has the Stop-server cleanup step it was suggesting.

Applied changes for the 3rd point. Regarding the other feedback:
1. The development team will be updated every time the user builds locally, so we can keep it as is for now.
2. We are not pushing `.lock` files, so there is no need to update them.
3. We updated the Flutter package, so it now requires a minimum Android version of 24 to work.

thias15:
Thanks for addressing points 3 and 4. One remaining thing:

**DEVELOPMENT_TEAM in `Runner.xcodeproj/project.pbxproj`**: this will cause confusing signing errors for external contributors building locally. Is there a specific reason to keep it hardcoded? If not, please set it to `""` — it's cleaner and each developer's Xcode will pick up their own signing configuration. If we ever need signed builds in CI, that can be injected as a secret.

PR #508, opened 2026-03-26
IOS CI workflow 

harishthakur52:
1. Installing cocoapods in system. 
2. Testing compatibilty with packages versions.


thias15 (approved):

thias15:
Fully tested and ready to merge?

harishthakur52:
Yes , tested and ready

thias15:
**Overall:** I'd second the Copilot review's three points — they're real issues even though the PR is approved:

1. **The workflow doesn't actually build the app.** The file header says "builds and tests the iOS app" but no `xcodebuild` step exists, so a broken Xcode project/scheme/signing config will still pass CI. Consider adding `xcodebuild -workspace OpenBot.xcworkspace -scheme OpenBot -destination 'generic/platform=iOS Simulator' build` (or similar) — otherwise this only validates that pods install, which is much narrower than the name implies.
2. **Pinning CocoaPods is worth doing.** `sudo gem install cocoapods` (no version) means a new CocoaPods release can spontaneously break CI without any code change.
3. **Deleting `Podfile.lock` makes CI non-deterministic.** If the goal is to ensure the lock matches reality, `pod install --deployment` (which fails when the lock is stale) is more useful than blowing it away. The current approach hides drift instead of catching it.

Happy to merge as a starting point if you'd rather iterate later, but (1) in particular reduces the value of this CI a lot.

harishthakur52:
For actually building the app on a simulator, we need to store our Apple team certificates and provisioning profile secrets and use them during the build process.

But as a starting point, I think we can merge this.

PR #516, opened 2026-05-21
Fixed: Blockly lights field toggle state resetting on reload.

hardikgarg02:
Issue: In Blockly Playground, When toggling the lights from "OFF" to "ON", reloading the page changes the value back to "OFF". The Blockly field state is not being saved.


thias15:
Good catch. The bug is clear: `!!"OFF"` is `true` (non-empty string), so the state always restored as ON.

Also worth noting in your description: removing `this.showEditor_()` from `init()` is a separate fix — that call was incorrectly opening the editor on every initialization.

thias15 (approved):

PR #529, opened 2026-07-28
Web controller fixes for both android and ios.

harishthakur52:
Fixes the web controller so it works reliably with both the Android and iOS robot apps.
 
- Reworked signaling in `server.js` and the iOS/Android clients so each web browser is correctly paired with its robot by room ID and client type, instead of broadcasting to everyone.

  - Fixed iOS bugs: video feed and keyboard control not working over the web controller, logging/network toggle and         camera-switch commands not being handled, inconsistent BLE wheel RPM display, and video feed rotating incorrectly on camera switch.
  
  - Fixed an Android camera-switch crash.


thias15:
Nice fixes here. A few questions before this can build/merge:

`WebSocketMessageHandler.swift` seems to be missing a closing brace on `toggleLogging()`, so `toggleNetwork()` and `quit()` end up nested inside it and the class never closes. Copilot's summary noticed this too ("currently has brace issue") but didn't flag it as a review comment. Wouldn't this fail to compile? (The iOS CI check only runs `pod install`, so it wouldn't catch it either.)

In `FreeRoamFragment.java`, the null-check on `connectionType.equals("USB")` doesn't carry over to `else if (connectionType.equals("Bluetooth"))` — could that NPE if `connectionType` is null? `LoggerFragment.java` and `RobotInfoFragment.java` got both branches checked.

Copilot also flagged the hardcoded IP in `NativeWebSocket.swift`, a `try!` in `sendRoomJoin()` that could crash on an encode failure, and `server.js`'s `rooms` map never removing empty entries — should these be fixed too?

Could you also add a PR description of what's fixed?

thias15:
Thanks, this addresses everything except the hardcoded IP. Quick fix for that: Android already has `ControllerConfig.getWebSignalingServerUrl()`/`setWebSignalingServerUrl()` (SharedPreferences-backed), just needs a settings field wired up. iOS would need the mirror: store it in `UserDefaults` instead of a hardcoded `let`, plus a matching settings field. Could you add those?

Separately (not blocking this PR): the old `controller/node-js` controller already advertises itself over mDNS (`_openbot-server._tcp.` in `dns.js`), and both apps already know how to discover it (`NsdService.java`, `Browser.swift`). `controller/web-server/server` doesn't do this yet. Worth doing the same here eventually, so users don't have to type an IP at all?

harishthakur52:
> Thanks, this addresses everything except the hardcoded IP. Quick fix for that: Android already has `ControllerConfig.getWebSignalingServerUrl()`/`setWebSignalingServerUrl()` (SharedPreferences-backed), just needs a settings field wired up. iOS would need the mirror: store it in `UserDefaults` instead of a hardcoded `let`, plus a matching settings field. Could you add those?
> 
> Separately (not blocking this PR): the old `controller/node-js` controller already advertises itself over mDNS (`_openbot-server._tcp.` in `dns.js`), and both apps already know how to discover it (`NsdService.java`, `Browser.swift`). `controller/web-server/server` doesn't do this yet. Worth doing the same here eventually, so users don't have to type an IP at all?

Thanks for the review! I've Addressed both points:

- Hardcoded signaling server URL: Replaced the hardcoded URL with a configurable Server URL setting on both Android and iOS (backed by SharedPreferences/UserDefaults). Verified end-to-end by connecting the Web Controller on both
platforms using the configured Server URL.

- mDNS: Not implemented at this time. Since the web server is intended to support remote URL's (e.g., wss://web-controller-server.onrender.com/ws),  mDNS would only benefit local network discovery and does not work to the current implementation scenerio.

url
url2
IMG_6665

thias15:
Confirmed all four previously flagged issues are actually fixed (checked the current file contents, not just the diff): the brace in `WebSocketMessageHandler.swift`, the null-check in `FreeRoamFragment.java`, `try!`/hardcoded IP in `NativeWebSocket.swift`, and the `rooms` map leak in `server.js` all look correct now.

A few small things from the "configurable server URL" commits:

1. `ControllerConfig.java` has a leftover commented-out line under the active `return`:
   ```java
   return get("web_signaling_server", "ws://192.168.1.45:8080/ws");
   //return get("web_signaling_server", "wss://web-controller-server.onrender.com/ws");
   ```
   Can this be removed?

2. The fallback default differs between platforms and both look like personal dev LAN IPs (iOS: `192.168.1.40`, Android: `192.168.1.45`). Worth making these consistent (or just using a placeholder-style value) since neither works for anyone else out of the box.

3. iOS's new "Web Signaling Server" label and placeholder are inline string literals, while everything else in `SettingsFragment.swift` goes through `Strings.swift`. Android's version correctly used `strings.xml`. Minor, but worth matching the file's existing convention.

One process note: PR #531 also rewrites `SettingsFragment.swift` significantly (Settings redesign), so merging both will hit a conflict there. Proposing we merge this one first, then rebase #531 on top and fix the conflict there, rather than the other way around.

thias15 (approved):
All previously flagged issues (unbalanced brace, missing null-check, force-try + hardcoded IP, rooms-map leak, and the follow-up cleanup on the configurable server URL) are fixed and verified against the current file contents. CI is green. Approving.

PR #531, opened 2026-08-04
IOS - App language support with flag icons and Settings redesign

hardikgarg02:
1. Added an in-app language picker (Settings → General → Language) with flag emoji for System Default, English, German, Spanish, French, Chinese, Korean, and Hindi.
2. Language switches live with a branded transition — no app restart needed.                                                                                             
3. Built full localization infrastructure from scratch (Localizable.strings for all 7 languages) — the app previously had none.                                          
4. Translated the entire app: Home, Settings, Profile, Projects, QR Scanner, Data Collection, Autopilot, Object Tracking, Robot Info, Navigation, and Playground screens.
5. Redesigned the Settings screen with section headers, icon-right rows, and the new Language row.                                                                       
6. Fixed Home screen tiles and tab bar titles (Projects/Profile) getting stuck in the original language after switching.                                                 
7. Fixed text truncation across multiple screens caused by label widths estimated from character count instead of real rendered width.                                   
8. Fixed off-center/truncated sign-in text and overlapping toggle labels ("Auto Mode", "Log Data") for longer translations.                                              
9. Made the Bluetooth Settings icon consistent with the one used on the Home screen.

### DIFF
--- a/ios/OpenBot/OpenBot.xcodeproj/project.pbxproj
+++ b/ios/OpenBot/OpenBot.xcodeproj/project.pbxproj
@@ -7,9 +7,11 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		0132784A30CAD8239CA1F112 /* en.lproj in Resources */ = {isa = PBXBuildFile; fileRef = 5022662904BF68D64CED43E6 /* en.lproj */; };
 		0619EECA2C24208B0050EB6A /* Modes.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0619EEC92C24208B0050EB6A /* Modes.swift */; };
 		06B219AD2ABD4D6100E65CEF /* SharedPreferences.swift in Sources */ = {isa = PBXBuildFile; fileRef = 06B219AC2ABD4D6100E65CEF /* SharedPreferences.swift */; };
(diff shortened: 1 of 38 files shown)


thias15:
Nice work — localization coverage is solid. Checked all 7 `Localizable.strings` files: key sets match exactly (163/163) and every `%@` format string is used consistently across languages. The `ModelManagementTable` fix (switching on a functional key instead of the localized dropdown label) was a good catch — that would've silently broken model filtering for non-English users otherwise.

One minor nit: `LocalizationManager.bundle` re-resolves the bundle path on every call, and every `Strings.*` access now routes through it. Probably fine at this scale, but caching it (invalidate on language change) would be a cheap win if it matters.

Was this rigorously tested (devices, languages, permission flows)?

hardikgarg02:
Applied localization bundle caching and fixed text overlapping issues for German (Deutsch) and Spanish. Tested across all languages and on the real robot.

thias15 (approved):
All previously flagged items are resolved: the `SettingsFragment.swift` merge-conflict overlap with #529's Web Signaling Server row is fixed (moved below Bluetooth, localized, keyboard-avoidance added), the `LocalizationManager.bundle` caching nit is fixed, and the two follow-up fixes (Speed/Input label overlap, camera orientation after language switch) look correct. CI is green, no conflicts with master. Approving.

PR #534, opened 2026-08-12
Update android robot package name, target API/Gradle, and localisations

hardikgarg02:
1. Updated Android robot app's package from org.openbot to org.openbot.app.robot.
2. Updated the robot module's target and compile SDK to API 36 and upgraded the Android Gradle Plugin to 8.9 ( required for Google Play Store listing )
3. Added missing translations for the Web Controller settings section and Web Signaling Server field (web_controller_category, web_signaling_server).


thias15:
One thing to confirm before merging: `google-services.json`'s `oauth_client` entries were edited to `package_name: "org.openbot.app.robot"` while keeping the same `certificate_hash` values. That's necessary but not sufficient — Google Sign-In needs an actual Android OAuth client registered in Firebase/Google Cloud Console for `(org.openbot.app.robot, that SHA-1)`, and editing the JSON locally doesn't create that registration. Has this been confirmed against the real console (and ideally tested with a live Google Sign-In on a build using the new applicationId)?

thias15:
Given #535 scrubbed the real Firebase credentials for iOS, could you do the same here? google-services.json still has the real client_id/certificate_hash/mobilesdk_app_id, and googleServices/README.md already has the matching "set up your Firebase project" walkthrough to add a dummy-data note to.

One thing first though: please confirm the org.openbot.app.robot OAuth client registration (from the comment above) actually works with a real Sign-In test before swapping to dummy values, since once it's dummied out there's nothing left to compare against the console.

thias15 (approved):
Everything's addressed, thanks. Good to merge from my side.

PR #535, opened 2026-08-12
IOS: updated bundle identifier to org.openbot.app.robot

hardikgarg02:


thias15:
GoogleService-Info.plist still declares `BUNDLE_ID = org.openbot`, but the app's `PRODUCT_BUNDLE_IDENTIFIER` is now `org.openbot.app.robot` — this file wasn't updated. Could you register `org.openbot.app.robot` as the iOS app in the Firebase console and swap in the fresh `GoogleService-Info.plist` it generates? Otherwise, wouldn't Firebase Auth/Google Sign-In (which the web-controller room-pairing in #529 depends on) fail to initialize correctly?

thias15:
Nice, the build step is a good addition, first time CI actually compiles the app rather than just resolving pods.

One thing still to fix: Info.plist (line 30) still has the real reversed client ID as the Google Sign-In URL scheme, so it wasn't dummied out with the rest. That also means sign-in will still fail for anyone following the README, since that scheme has to match the REVERSED_CLIENT_ID of their own Firebase project. Could you dummy it out and add that step to the README?

thias15 (approved):
Everything's addressed, thanks. Good to merge from my side.

Issue #447, opened 2024-08-16
the layer MLP has never been called and thus has no defined output


sparta-2024:
**Describe the bug**
When using tensorflow 2.16 or 2.17 to do train ,there is  an  error occurred:the layer MLP has never been called and thus has no defined output.
**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [e.g. Windows, Linux, iOS]

**Smartphone (please complete the following information):**
 - Device: [e.g. Samsung S20]
 - OS: [e.g. Android 11]
 - Browser [e.g. stock browser, chrome]
 - Version [e.g. 0.1.1, master]

**Additional context**

sparta-2024:
models.py file should be added the following code:
import os
os.environ["TF_USE_LEGACY_KERAS"] = "1"

isha382:
hi @sparta-2024, the issue is of the version of tensorflow you can try with 2.9.0 version which is preferred version and can refer following documentation [policy](https://github.com/3dwesupport/OpenBot/blob/master/policy/README.md) for more information.

thias15:
Opened PR #514 with the legacy-Keras flag (set in `policy/openbot/__init__.py` so it applies before any submodule imports TensorFlow). Once that lands, training should work on TF >= 2.16 without the MLP error and without forcing a downgrade. Thanks @sparta-2024 for finding the workaround.

PR #537, opened 2026-08-21
test(node-js): add unit tests for controller command logic; fix DriveValue.read rounding

Arvin7liu:
## Summary

The node-js web controller has no automated tests for its server-side command logic, even though that logic computes the exact left/right motor commands sent to the robot. This PR adds dependency-free unit tests and fixes a latent rounding bug found while writing them.

- **New tests** (`controller/node-js/test/commands.test.js`, run via Node's built-in `node:test`):
  - `CommandHandler` drive mapping for `goForward` / `goBackward` / `forwardLeft` / `forwardRight` / `rotateLeft` / `rotateRight` / `reset`.
  - `DriveCommandReducer` de-duplication (identical consecutive drive commands are not re-sent).
  - `CommandHandler.sendCommand` raw-command forwarding.
  - `DriveValue` clamping to [-1, 1] and `read()` precision.
- **Bug fix**: `DriveValue.read()` used `Math.round(value, 3)`. `Math.round` only accepts one argument, so the `3` was ignored and the method returned an **integer** instead of a 3-decimal value. Changed to `Math.round(value * 1000) / 1000`. `DriveValue` is now exported so the tests can reach it. Note: `read()` isn't currently called anywhere in `server/` or `client/` (the drive-command path uses the raw `.min()`/`.max()`/`.write()` values directly), so this fix doesn't change any live behavior today, it's a correctness fix to a previously-unused method, now exercised by this PR's own tests.
- Added a `"test": "node --test"` script to `controller/node-js/package.json` (zero new dependencies; runs on Node 18+).

## Why this matters

The `CommandHandler` drive mapping tested here is the code that computes the actual left/right motor commands sent to the robot, so a regression there would silently send wrong motor commands to a physical robot. The new tests lock in that expected behavior and run in CI-friendly, dependency-free fashion.

## Test plan

- [x] `cd controller/node-js && npm test` → 10/10 passing on Node 22 (also verified with `node --check` on all `server/*.js` and `client/*.js`).
- [x] Regression check: reverting the `read()` fix makes the `DriveValue` test fail (proving the test actually guards the fix).

## Notes

This is unrelated to #447 / PR #514 (the TF/Keras policy-training issue), which is handled separately.

🤖 Generated with WorkBuddy

### DIFF
--- a/controller/node-js/package.json
+++ b/controller/node-js/package.json
@@ -10,7 +10,8 @@
     "dev:server": "npm run start --prefix server",
     "dev:start-client": "vite --port 8081 --host",
     "build": "vite build",
-    "lint": "eslint server/*.js --fix && eslint client/*.js --fix"
+    "lint": "eslint server/*.js --fix && eslint client/*.js --fix",
+    "test": "node --test"
   },
   "author": "Ivo Zivkov",
   "devDependencies": {
--- a/controller/node-js/server/commands.js
+++ b/controller/node-js/server/commands.js
@@ -42,7 +42,7 @@ function DriveValue () {
   }
 
   this.read = () => {
-    return Math.round(value, 3)
+    return Math.round(value * 1000) / 1000
   }
 }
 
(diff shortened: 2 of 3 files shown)


thias15 (approved):
Verified all test assertions by hand against the actual CommandHandler/DriveValue logic; they're correct. CI is green. Approving.

PR #538, opened 2026-08-25
Updated Android releases.yml for keystore and google-services secrets

hardikgarg02:
1. Updated Android releases.yml for keystore and google-services secrets
2. Fixed the edge-to-edge screen layout issue on Android devices running Android 15 or above.
3. Fixed issue for app crash in 16 KB devices.


thias15:
`google-services.json` is decoded to both `robot/google-services.json` and `robot/src/main/assets/google-services.json`. Is the assets copy needed at runtime? If so, could you clarify where it’s consumed?

harishthakur52:
> `google-services.json` is decoded to both `robot/google-services.json` and `robot/src/main/assets/google-services.json`. Is the assets copy needed at runtime? If so, could you clarify where it’s consumed?

 Yes, the assets copy is needed at runtime. BarCodeScannerFragment.readJsonFile() (android/robot/src/main/java/org/openbot/app/robot/projects/BarCodeScannerFragment.java) opens google-services.json directly from context.getAssets() and parses it in getCurrentKeyFromJson() to extract the API key used for [barcode scanning / drive file access].

thias15:
Thanks for clarifying why the assets copy is needed.

I also noticed this PR includes the WebRTC library replacement, TensorFlow Lite → LiteRT migration, ARCore upgrade, and main screen layout changes. Could you confirm what testing has been done for these, specifically:

- Building and signing both release APKs through the updated workflow.
- Video streaming between the robot and controllers.
- Model inference, including GPU acceleration, and AR functionality.
- Screen layout and compatibility on 16 KB page-size devices.

Gradle checks are passing, but these runtime changes need validation beyond compilation. If they can be handled independently, could we move them into a separate PR and keep this one focused on release signing?

harishthakur52:
We have tested all the above points you mentioned for both the Robot and Controller apps, and everything is working fine. We also tested them on Android 11 and Android 16 devices. So, I think we can merge them.

We did notice one issue while testing ARCore. The fix has been added in the PR #542 .

